PR #29, opened 2020-07-01
feat: Courier Sync

sdsantos:
Initial version of the courier sync. 

What it does:
- Detects the correct IP address of the Courier device
- Connects to the courier and fakes a regular sync


app/src/main/java/tech/relaycorp/gateway/domain/StoreParcel.kt
```diff
@@ -0,0 +1,80 @@
...
+        } catch (exc: RAMFException) {
+            logger.warning("Invalid cargo received: ${exc.message}")
+            return Result.Error.Invalid
+        }
+
+        return storeMessage(parcel, parcelBytes)
+    }
+
+    private suspend fun storeMessage(
+        message: RAMFMessage,
```
gnarea, line 41: We probably shouldn't try to make this a generic method: it can only work with parcels.

app/src/main/java/tech/relaycorp/gateway/data/model/ParcelDirection.kt
```diff
@@ -0,0 +1,11 @@
+package tech.relaycorp.gateway.data.model
+
+enum class ParcelDirection(val value: String) {
+    Incoming("incoming"), Outgoing("outgoing");
```
gnarea, line 4: I'd suggest to avoid the terms "incoming" and "outgoing" as that's ambiguous in a broker like this gateway, where its responsibility is to route data.

This is the terminology I used on the public gateway to avoid ambiguity:

- "(Private) gateway-bound parcels": Parcels bound for a private gateway. Those parcels will be encapsulated in cargo if a courier picks them up, or sent as is when using PoWebSockets.
- "Internet-bound parcels": Parcels bound for an endpoint on the Internet. Behind the scenes, that endpoint could be a service (e.g., `https://api.twitter.com/relaynet`) or another gateway (e.g., `https://public-gateway.relaycorp-competitor.com`) if the recipient of the parcel is a private endpoint.

However, I don't think they'd work here because things will be more complex in private gateways:

- One optimisation we need to do after this project is to deliver parcels bound for public Internet endpoints directly, skipping the public gateway, when the Internet is available and the public endpoint is not blocked by the ISP. We'll only send the parcel to our public gateway when we can't deliver it directly.
- We're also going to support other underlying networks in the medium term, not just the Internet and courier-based sneakernets. For example, we'll eventually support Bluetooth-based meshnets, where parcels could be broadcast to all gateways on the meshnet or sent to a specific gateway -- and those parcels will never be sent to the Internet or a courier.

For the reasons above, I'd suggest the following terms for now (but I'm very much open to suggestions):

- "Local endpoint-bound parcels".
- "Internet endpoint-bound parcels".

Thoughts?

app/src/main/java/tech/relaycorp/gateway/data/repos/ParcelRepository.kt
```diff
@@ -0,0 +1,27 @@
...
+    suspend fun insert(message: StoredParcel)
+
+    @Delete
+    suspend fun delete(message: StoredParcel)
+
+    @Query("SELECT * FROM Parcel")
+    fun observeAll(): Flow<List<StoredParcel>>
+
+    @Query("SELECT COALESCE(SUM(Parcel.size), 0) FROM Parcel")
+    fun observeTotalSize(): Flow<StorageSize>
```
gnarea, line 23: I guess methods like this will depend on the destination of the parcels. For example, in the screen where we should the size of the data to be sent to couriers, we'd want to exclude the data that hasn't yet been delivered to local endpoints.

app/src/main/java/tech/relaycorp/gateway/data/model/ParcelDirection.kt
gnarea: ... In other words, I don't think we should be thinking about the _direction_ of the parcel, but I think we should be thinking about the _location of its target endpoint_:

- Local network.
- Internet.
- Meshnet (outside current scope).
- Etc.

app/schemas/tech.relaycorp.gateway.data.database.AppDatabase/1.json
```diff
@@ -2,33 +2,87 @@
...
+            "columnName": "size",
             "affinity": "INTEGER",
             "notNull": true
           }
         ],
         "primaryKey": {
           "columnNames": [
-            "id"
+            "senderAddress",
+            "messageId"
```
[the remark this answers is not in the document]
sdsantos: I thought the rule was that they were unique per sender. I imagine two different senders could have the same message id for the same recipient no, cause that's harder to enforce?

app/schemas/tech.relaycorp.gateway.data.database.AppDatabase/1.json
```diff
@@ -2,33 +2,87 @@
...
+            "columnName": "messageId",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "creationTimeUtc",
+            "columnName": "creationTimeUtc",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
```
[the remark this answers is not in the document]
sdsantos: You mentioned ordering parcels inside a cargo by the creation time of the parcel, that's why I added it.

app/schemas/tech.relaycorp.gateway.data.database.AppDatabase/1.json
```diff
@@ -2,33 +2,87 @@
   "formatVersion": 1,
   "database": {
     "version": 1,
-    "identityHash": "f4d34def0a467a21b391ecb53e2175d2",
+    "identityHash": "9ddb31ac692f375edbe5f934ce68bf51",
     "entities": [
       {
-        "tableName": "Message",
-        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`id` INTEGER NOT NULL, PRIMARY KEY(`id`))",
+        "tableName": "Parcel",
```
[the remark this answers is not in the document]
sdsantos: We can always migrate the current database schema, once we want to add more functionality. The initial schema does not need to be the final one.

app/src/main/java/tech/relaycorp/gateway/data/model/MessageAddress.kt
```diff
@@ -0,0 +1,37 @@
...
+        }
+    val value
+        get() = when (this) {
+            is PublicMessageAddress -> publicValue
+            is PrivateMessageAddress -> privateValue
+        }
+
+    companion object {
+        fun of(value: String) =
+            if (value.contains(":")) {
```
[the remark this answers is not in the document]
sdsantos: It would be nice if the core library had some Address functionality for the apps to lean on. I basically just copied this class over from the Courier app.

app/schemas/tech.relaycorp.gateway.data.database.AppDatabase/1.json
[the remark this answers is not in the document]
gnarea: It's unique per channel -- That is, the parcel id should be unique for any sender/recipient combination.

In other words, if we want to represent the primary key as a composite key, the underlying keys would be: sender's private address address, recipient's address and parcel id.

app/schemas/tech.relaycorp.gateway.data.database.AppDatabase/1.json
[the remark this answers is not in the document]
gnarea: Of course! Sorry, disregard my comment.

app/src/main/java/tech/relaycorp/gateway/domain/StoreParcel.kt
sdsantos: This `StoreParcel` is still WIP, I copied the StoreMessage over from the Courier app and renamed things. But this will be rewritten when it actually becomes used.

app/schemas/tech.relaycorp.gateway.data.database.AppDatabase/1.json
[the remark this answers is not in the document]
gnarea: Agreed!

app/src/main/java/tech/relaycorp/gateway/domain/StoreParcel.kt
sdsantos: I can remove it actually from this PR.

app/src/main/java/tech/relaycorp/gateway/data/model/MessageAddress.kt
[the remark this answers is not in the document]
gnarea: Agreed! I'll create a PR when https://github.com/relaycorp/relaynet-jvm/issues/62 is done

app/src/main/java/tech/relaycorp/gateway/data/model/ParcelDirection.kt
sdsantos: Ok, definitely more confusing to think of all those scenarios right now 😅

So a kind of `recipientLocation` or `targetLocation`?

app/src/main/java/tech/relaycorp/gateway/data/repos/ParcelRepository.kt
sdsantos: Right, this method was copied over, but it's not being used at the moment. I can remove it, and only add it when we have an actual usage for it.

app/src/main/java/tech/relaycorp/gateway/data/model/ParcelDirection.kt
gnarea: I think so! I'd prefer the former because it has the word "recipient" in it.

app/src/main/java/tech/relaycorp/gateway/background/CogRPC.kt
```diff
@@ -0,0 +1,5 @@
+package tech.relaycorp.gateway.background
+
+object CogRPC {
+    const val PORT = 21473
```
gnarea, line 4: Heads up: I'll move this to the core lib: https://github.com/relaycorp/relaynet-jvm/issues/63

Issue #19, opened 2019-08-12
Implement cargo collection


gnarea:
Using https://github.com/relaycorp/relaynet-cosocket-java/issues/7

The cargo must be delivered after the relayer has produced a valid Cargo Collection Authorization.

TODO:

- [x] Use the private gateway's certificate when validating cargo and parcels coming from the Internet/couriers.

PR #37, opened 2020-07-07
feat: Courier collection cargo and parcel operations

sdsantos:
Implementing cargo and parcel operations related with the Courier cargo collection, minus generating the CCA.


app/src/main/java/tech/relaycorp/gateway/domain/courier/ProcessCargo.kt
```diff
@@ -1,10 +1,26 @@
...
 
-    // TODO: implement
-    suspend fun process() {}
+    suspend fun process() {
+        val cargoes = cargoStorage.list()
+        cargoes.forEach { cargoStream ->
+            val cargo = Cargo.deserialize(cargoStream().readBytesAndClose())
+            readParcelsFromCargo
+                .read(cargo)
+                .forEach { storeParcel.store(it) }
```
gnarea, line 22: Keep in mind a cargo may contain other things besides parcels, such as PCAs. Happy to keep PCAs outside the scope of this PR though, I just wanted to remind you just in case.

app/src/main/java/tech/relaycorp/gateway/data/disk/CargoStorage.kt
```diff
@@ -0,0 +1,51 @@
...
+        val cargoBytes = cargoStream.readBytesAndClose()
+        val cargo = try {
+            Cargo.deserialize(cargoBytes)
+        } catch (exc: RAMFException) {
+            logger.warning("Malformed cargo received: ${exc.message}")
+            return Operation.Error(MalformedCargoException())
+        }
+
+        try {
+            cargo.validate()
```
gnarea, line 26: FYI: I added a TODO to https://github.com/relaycorp/relaynet-gateway-android/issues/19 to make sure to pass the gateway's certificate to this method, once we have the code to generate the gateway's certificate.

app/src/main/java/tech/relaycorp/gateway/common/Operation.kt
```diff
@@ -0,0 +1,9 @@
+package tech.relaycorp.gateway.common
+
+sealed class Operation<out T> {
+    data class Success<T>(val result: T) : Operation<T>()
+    data class Error<T>(val throwable: Throwable) : Operation<T>()
+}
+
+val <T> Operation<T>.isSuccessful get() = this is Operation.Success<T>
+val <T> Operation<T>.isError get() = this is Operation.Error<T>
```
gnarea, line 9: Out of curiosity as someone new to Kotlin/Android, why use this class vs simply raising exceptions when things go wrong?

app/src/main/java/tech/relaycorp/gateway/data/disk/DiskOperations.kt
```diff
@@ -0,0 +1,74 @@
...
+            getOrCreateDir(folder)
+                .listFiles()
+                ?.map { it::inputStream }
+                ?: emptyList()
+        }
+
+    @Throws(ParcelDataNotFoundException::class)
+    suspend fun readMessage(folder: String, path: String): (() -> InputStream) {
+        val file = File(getOrCreateDir(folder), path)
+        if (!file.exists()) throw ParcelDataNotFoundException("Parcel data not found on path '$path'")
```
gnarea, line 37: I'd suggest to be consistent: Is this class/method dealing with messages in general or parcels specifically?

app/src/main/java/tech/relaycorp/gateway/data/disk/DiskOperations.kt
```diff
@@ -0,0 +1,74 @@
...
+    private suspend fun getOrCreateDir(folder: String) =
+        withContext(Dispatchers.IO) {
+            File(context.filesDir, folder).also {
+                if (!it.exists()) it.mkdir()
+            }
+        }
+
+    private suspend fun createUniqueFile(folder: String, prefix: String) =
+        withContext(Dispatchers.IO) {
+            val parcelsDir = getOrCreateDir(folder)
```
gnarea, line 70: Same comment about parcels vs messages applies here

app/src/main/java/tech/relaycorp/gateway/domain/courier/DeliverParcel.kt
```diff
@@ -0,0 +1,17 @@
+package tech.relaycorp.gateway.domain.courier
+
+import tech.relaycorp.gateway.common.Operation
+import tech.relaycorp.gateway.data.model.MessageAddress
+import java.io.InputStream
+import javax.inject.Inject
+
+class DeliverParcel
```
gnarea, line 8: This will do the parcel delivery to a local endpoint, right? In that case, I think it should live outside `tech.relaycorp.gateway.domain.courier` because we'd need this code once we're able to sync over the Internet directly.

app/src/main/java/tech/relaycorp/gateway/domain/courier/ProcessCargo.kt
sdsantos: Right, forgot about that. I'll add a `TODO` about it.

app/src/main/java/tech/relaycorp/gateway/common/Operation.kt
sdsantos: It's a good question, and I thought about it myself too. This Operation pattern makes it more clear that the error case should be handled. The sealed class just helps for the kotlin `when`. This is the reactive programming pattern, not far away for the `Either` pattern: https://www.ibm.com/developerworks/library/j-ft13/index.html

But in practice, there's not a huge difference.

app/src/main/java/tech/relaycorp/gateway/data/disk/DiskOperations.kt
sdsantos: Oh right, that's a mistake, it's a generic helper class, so messages.

app/src/main/java/tech/relaycorp/gateway/data/disk/DiskOperations.kt
sdsantos: My bad.

app/src/main/java/tech/relaycorp/gateway/domain/courier/DeliverParcel.kt
sdsantos: Right, well spotted.

app/src/main/java/tech/relaycorp/gateway/common/Operation.kt
gnarea: :+1: Thanks for the explanation!

Issue #1, opened 2019-02-21
Write up README


gnarea:
Including but not limited to:

- [x] Install instructions.
- [x] Pre-registration.
- [x] Write up security and privacy considerations:
  - [x] Keystore on Android 5: https://issuetracker.google.com/issues/132325342#comment29. Mention that the threat can be mitigated with full-disk encryption.
  - [x] Gateway has to store the `applicationId` of each endpoint app in order to wake it up when a new parcel is received.

PR #45, opened 2020-07-13
feat: Generate CCA

sdsantos:


app/build.gradle
```diff
@@ -90,6 +94,7 @@ dependencies {
     implementation 'androidx.cardview:cardview:1.0.0'
     implementation 'androidx.lifecycle:lifecycle-viewmodel-ktx:2.2.0'
     implementation 'androidx.lifecycle:lifecycle-runtime-ktx:2.2.0'
+    implementation 'androidx.security:security-crypto:1.1.0-alpha01'
```
gnarea, line 97: Can you please add a TODO to the README to elaborate on the risks on API 21-22? Just a quick note to elaborate on it when we write the documentation.

app/src/main/java/tech/relaycorp/gateway/domain/LocalConfig.kt
```diff
@@ -0,0 +1,73 @@
...
+            ?: generateGatewayCertificate(getKeyPair())
+                .also {
+                    sensitiveStore.store(CERTIFICATE_FILE_NAME, it.serialize())
+                }
+
+    private fun generateKeyPair(): KeyPair {
+        val keyGen = KeyPairGenerator.getInstance(KEY_ALGORITHM)
+        keyGen.initialize(RSA_KEY_MODULUS)
+        return keyGen.generateKeyPair()
+    }
```
gnarea, line 42: I'd suggest to use the [`generateRSAKeyPair` function from the Relaynet JVM lib](https://github.com/relaycorp/relaynet-jvm/blob/da8eabbb2335077704d84f27fdf91890053bbba5/src/main/kotlin/tech/relaycorp/relaynet/wrappers/Keys.kt#L16), as it does this exactly, but also makes sure that only RSA keys are generated (using other types of keys in unsupported in Relaynet).

app/src/main/java/tech/relaycorp/gateway/domain/LocalConfig.kt
```diff
@@ -0,0 +1,73 @@
...
+        return generator.generatePrivate(privateKeySpec)
+    }
+
+    private fun PrivateKey.toKeyPair(): KeyPair {
+        val publicKeySpec =
+            (this as RSAPrivateCrtKey).run { RSAPublicKeySpec(modulus, publicExponent) }
+        val keyFactory = KeyFactory.getInstance("RSA")
+        val publicKey = keyFactory.generatePublic(publicKeySpec)
+        return KeyPair(publicKey, this)
+    }
```
gnarea, line 63: FYI: I created an issue in the Relaynet lib to move the implementation there, since we'll have to (de)serialise keys in the Android endpoint too. https://github.com/relaycorp/relaynet-jvm/issues/66

app/src/main/java/tech/relaycorp/gateway/domain/courier/GenerateCCA.kt
```diff
@@ -1,15 +1,24 @@
...
 
-    // TODO: implementation
-    fun generate() =
-        CargoCollectionAuthorization.deserialize(ByteArray(0))
+    suspend fun generate() =
+        CargoCollectionAuthorization(
+            recipientAddress = publicGatewayPreferences.getAddress().first(),
+            payload = "".toByteArray(),
+            senderCertificate = localConfig.getCertificate()
+        )
```
gnarea, line 20: Sorry, I forgot to specify this, but can you please set a TTL of 14 days on the CCA constructor? (The TTL is an integer representing the number of seconds since the creation time of the CCA -- the current time, in this case)

app/src/main/java/tech/relaycorp/gateway/domain/courier/GenerateCCA.kt
```diff
@@ -1,15 +1,24 @@
...
 
-    // TODO: implementation
-    fun generateByteArray() = ByteArray(0)
+    suspend fun generateByteArray() =
+        generate().serialize(localConfig.getKeyPair().private)
```
gnarea, line 23: Can you please add test(s) for this? I'd suggest using `CargoCollectionAuthorization.deserialize()` too

app/build.gradle
gnarea: Never mind, I added a TODO to https://github.com/relaycorp/relaynet-gateway-android/issues/1

Issue #16, opened 2019-08-12
Implement parcel collection


gnarea:
- [x] Implement nonce generation and verification using CMS SignedData values (plaintext attached, signer cert detached).
- [x] Pass private gateway's certificate to `parcel.validate()` when the parcel is coming from the public gateway. This requires implementing that argument in `RAMFMessage.validate()`, like its counterpart in the JS lib.
- [x] Handshake: Refuse signer certs not issued by gateway.
- [x] Call cert.getCertificartePath()
- [x] Actually implement the parcel collection code. (#92)
- [x] Write functional/instrumentation test (requires integrating https://github.com/relaycorp/relaynet-poweb-jvm/pull/29 in the test)

PR #53, opened 2020-07-20
feat: Nonce generation service

sdsantos:


app/schemas/tech.relaycorp.gateway.data.database.AppDatabase/1.json
```diff
@@ -2,8 +2,35 @@
...
+            "fieldPath": "address",
+            "columnName": "address",
+            "affinity": "TEXT",
+            "notNull": true
+          }
+        ],
+        "primaryKey": {
+          "columnNames": [
+            "address",
+            "applicationId"
```
gnarea, line 27: I think the applicationId shouldn't be part of the PK: Whilst I don't anticipate it'd be a frequent thing, it should be possible to migrate endpoints to another app.

Once we start working on the endpoint implementation, you'll find that each Relaynet service defines the _contract_ to be adhered to by each endpoint participating in the service. Amongst other things, that means Relaynet or its gateways don't care about apps, but endpoints, and -- in turn -- that means that users can migrate their data (inc. endpoint crypto keys) to other compatible apps. 

Say WhatsApp and Signal share the same Relaynet service: People would be able to migrate between those apps without disrupting their communication. However, the actual migration is outside the scope of Relaynet, but it should still support reassigning an endpoint to another app. The team defining the Relaynet-based messaging service should define the export/import process.

app/src/main/AndroidManifest.xml
```diff
@@ -7,6 +7,10 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
 
+    <permission
+        android:name="tech.relaycorp.relaynet.GATEWAY_SYNC"
```
gnarea, line 11: Shouldn't this be called `tech.relaycorp.gateway.<whatever>` for consistency with the applicationId / package name?

app/src/main/java/tech/relaycorp/gateway/domain/endpoint/GenerateEndpointNounce.kt
```diff
@@ -0,0 +1,10 @@
+package tech.relaycorp.gateway.domain.endpoint
+
+import javax.inject.Inject
+
+class GenerateEndpointNounce
+@Inject constructor() {
+
+    // TODO: implementation
```
gnarea, line 8: FYI: I added a TODO to #16

app/src/androidTest/java/tech/relaycorp/gateway/background/endpoint/NounceRequestServiceTest.kt
```diff
@@ -0,0 +1,48 @@
...
+import android.os.Message
+import android.os.Messenger
+import androidx.test.core.app.ApplicationProvider.getApplicationContext
+import androidx.test.rule.ServiceTestRule
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotNull
+import org.junit.Rule
+import org.junit.Test
+
+class NounceRequestServiceTest {
```
gnarea, line 16: Can you please add a test to make that nothing breaks, hangs or times out when sending an invalid request? (I.e., one without `NounceRequestService.NOUNCE_REQUEST`)

app/schemas/tech.relaycorp.gateway.data.database.AppDatabase/1.json
sdsantos: But how can we notify an app that we got that for them then? I was expecting to use the `applicationId` for that. Without it, the only alternative I had in mind is waking up all relaynet apps with a parameter of the endpoint address we actually wanted to notify.

app/src/main/AndroidManifest.xml
sdsantos: Could be, as it could also be global, but I can set it to `tech.relaycorp.relaynet.gateway.SYNC`

app/schemas/tech.relaycorp.gateway.data.database.AppDatabase/1.json
gnarea: You'd still store the application Id. It just wouldn't be part of the primary key.

app/src/androidTest/java/tech/relaycorp/gateway/background/endpoint/NounceRequestServiceTest.kt
sdsantos: Ok.

app/src/main/java/tech/relaycorp/gateway/data/model/LocalEndpoint.kt
```diff
@@ -1,12 +1,15 @@
 package tech.relaycorp.gateway.data.model
 
 import androidx.room.Entity
+import androidx.room.Index
+import androidx.room.PrimaryKey
 
 @Entity(
     tableName = "Endpoint",
-    primaryKeys = ["address", "applicationId"]
+    indices = [Index(value = ["applicationId"], unique = false)]
```
gnarea, line 9: Why isn't it a PK anymore? And why `unique=false`?

app/src/main/AndroidManifest.xml
gnarea: Shouldn't `gateway` come before `relaynet` as in `tech.relaycorp.gateway.relaynet.SYNC`, since `tech.relaycorp.gateway` is the applicationId and package name?

app/src/main/java/tech/relaycorp/gateway/data/model/LocalEndpoint.kt
sdsantos: If there's only one PK, we can use the annotation `@PrimaryKey` straight on the `address` field. And I marked the `applicationId` and non-unique because I believe we discussed that one app could have multiple addresses.

app/src/main/AndroidManifest.xml
sdsantos: I don't know if there's a convention for this, but it does make more sense.

app/src/main/java/tech/relaycorp/gateway/data/model/LocalEndpoint.kt
gnarea: > If there's only one PK, we can use the annotation @PrimaryKey straight on the address field

Indeed, there'd only be one PK.

> And I marked the applicationId and non-unique because I believe we discussed that one app could have multiple addresses

My bad! Yes, that's correct.

PR #54, opened 2020-07-20
feat: Cargo message processing

sdsantos:


app/schemas/tech.relaycorp.gateway.data.database.AppDatabase/1.json
```diff
@@ -119,12 +119,46 @@
           }
         ],
         "foreignKeys": []
+      },
+      {
+        "tableName": "ParcelAck",
```
gnarea, line 124: I think `ParcelCollection` would be a more accurate name, since these records will exist before any their ACKs are generated/sent. Also, in the future we're likely to store metadata about the collection, like the address of the peer gateway.

app/schemas/tech.relaycorp.gateway.data.database.AppDatabase/1.json
```diff
@@ -119,12 +119,46 @@
...
+        "fields": [
+          {
+            "fieldPath": "recipientAddress",
+            "columnName": "recipientAddress",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "senderAddress",
+            "columnName": "senderAddress",
```
gnarea, line 135: I think this should be `senderPrivateAddress` to be more specific -- it should never be set to a public address.

app/schemas/tech.relaycorp.gateway.data.database.AppDatabase/1.json
```diff
@@ -119,12 +119,46 @@
           }
         ],
         "foreignKeys": []
+      },
+      {
+        "tableName": "ParcelAck",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`recipientAddress` TEXT NOT NULL, `senderAddress` TEXT NOT NULL, `messageId` TEXT NOT NULL, PRIMARY KEY(`recipientAddress`, `senderAddress`, `messageId`))",
+        "fields": [
```
gnarea, line 126: This is missing the expiry date of the parcel, which is important here: Records with an expiry date in the past should be automatically removed.

app/src/main/java/tech/relaycorp/gateway/domain/StoreParcel.kt
```diff
@@ -25,10 +25,15 @@ class StoreParcel
...
+
+    @Throws(MalformedParcelException::class, InvalidParcelException::class)
+    suspend fun store(
+        parcelData: ByteArray,
+        recipientLocation: RecipientLocation
     ): StoredParcel {
-        val parcelBytes = parcelStream.readBytesAndClose()
         val parcel = try {
-            Parcel.deserialize(parcelBytes)
+            Parcel.deserialize(parcelData)
```
gnarea, line 36: (This comment applies to line 43, which is outside the diff)

We should enforce that `parcel.isRecipientAddressPrivate == true` when receiving parcels from the public endpoint. We shouldn't do that when the parcel originated in a private endpoint on the phone.

FYI: We have to pass the private gateway's certificate to `parcel.validate()` when the parcel is coming from the public gateway. We shouldn't do that when the parcel originated in a private endpoint on the phone. However, I forgot to implement this optional argument in `RAMFMessage.validate()`, so I've just added a TODO to #16 to fix this.

In other words, the two things above will depend on the value of `recipientLocation`.

app/src/main/java/tech/relaycorp/gateway/domain/courier/ProcessCargo.kt
```diff
@@ -1,30 +1,67 @@
...
+    private val readMessagesFromCargo: ReadMessagesFromCargo,
+    private val storeParcel: StoreParcel,
+    private val storeParcelAck: StoreParcelAck,
+    private val deleteParcel: DeleteParcel
 ) {
 
     suspend fun process() {
         val cargoes = cargoStorage.list()
         cargoes.forEach { cargoStream ->
             val cargo = Cargo.deserialize(cargoStream().readBytesAndClose())
```
gnarea, line 30: We should do `cargo.validate()` before any further processing.

app/src/sharedTest/java/tech/relaycorp/gateway/test/factory/ParcelFactory.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    fun buildSerialized() =
+        build().serialize(keyPair.private)
+
+    private val keyPair by lazy {
+        generateRSAKeyPair()
+    }
+
+    private val certificate by lazy {
+        issueGatewayCertificate(
```
gnarea, line 24: I'd suggest to use `issueEndpointGateway` here instead, since parcels are signed by endpoints. Future tests may break if we don't use the right type of certificate.

app/src/test/java/tech/relaycorp/gateway/domain/courier/ProcessCargoTest.kt
```diff
@@ -0,0 +1,72 @@
...
+import tech.relaycorp.gateway.domain.DeleteParcel
+import tech.relaycorp.gateway.domain.StoreParcel
+import tech.relaycorp.gateway.domain.StoreParcelAck
+import tech.relaycorp.gateway.test.factory.CargoFactory
+import tech.relaycorp.gateway.test.factory.ParcelCollectionAckFactory
+import tech.relaycorp.gateway.test.factory.ParcelFactory
+import tech.relaycorp.gateway.test.factory.StoredParcelFactory
+import tech.relaycorp.relaynet.messages.payloads.CargoMessage
+
+class ProcessCargoTest {
```
gnarea, line 23: I'd suggest to add tests for:

- Invalid/unsupported messages in cargoes.
- Malformed parcels.
- Well-formed but invalid parcels.

app/schemas/tech.relaycorp.gateway.data.database.AppDatabase/1.json
sdsantos: Can't we get it from the Parcel table by relationship?

app/src/main/java/tech/relaycorp/gateway/domain/courier/ProcessCargo.kt
sdsantos: That was already done before storing the Cargo. We can do it again here, but it would be a bit wasteful.

gnarea: Sorry, I didn't realise. In that case disregard my comment.

app/schemas/tech.relaycorp.gateway.data.database.AppDatabase/1.json
gnarea: I don't think so: The parcel could've been delivered to a local endpoint and subsequently deleted, before we get to generate/send PCAs for the courier.

app/schemas/tech.relaycorp.gateway.data.database.AppDatabase/1.json
sdsantos: Here the sender is coming from the outside, shouldn't it be public or private? Probably the recipient is the one that can only be private?

gnarea: > Here the sender is coming from the outside, shouldn't it be public or private?

The sender could indeed be a public or private endpoint. However, we can't tell from the outside whether the sender has a public address: We can only see its private address.

That's because, in a RAMF message, there's a metadata field for the recipient's private or public address, but no counterpart for the sender. Instead, the sender's private address is _inferred_ from the sender's certificate, which is contained in the digital signature -- Since private addresses are based on the public key in the certificate, we get it by calling `certificate.subjectPrivateAddress`.

The main reason for inferring the private address of the sender (instead of explicitly stating its private/public address in the metadata) is that the sender's public address (if any) is never used for routing or authorisation purposes.

> Probably the recipient is the one that can only be private?

Syntactically, the address could be either private or public. But because this is a private gateway, its public gateway should only be sending parcels bound for private endpoints.

However, we still have to send PCAs for invalid parcels so that the public gateway won't send them again. That means we may end up recording parcel collections for parcels bound for public endpoints -- even though those parcels are invalid and should be refused. I can think of two unlikely reasons why a public gateway would do that:

- A bug. The public gateway I implemented would refuse those invalid parcels -- But there could be a bug in the implementation.
- A rogue public gateway.

PR #55, opened 2020-07-23
feat: Generate cargo

sdsantos:


app/src/main/java/tech/relaycorp/gateway/common/DateUtils.kt
```diff
@@ -0,0 +1,8 @@
+package tech.relaycorp.gateway.common
+
+import java.time.ZoneId
+import java.time.ZonedDateTime
+import java.util.Date
+
+fun Date.toZonedDateTime(): ZonedDateTime = ZonedDateTime.ofInstant(toInstant(), ZoneId.of("UTC"))
```
gnarea, line 7: Shouldn't this be `ZoneId.systemDefault()` instead of `ZoneId.of("UTC")`? I've been bitten by this before and I only realised when my tests broke in the transition from GMT to BST :smile:  But maybe Android is different. Would be good to check.

I suspect we may have to do something like:

```
ZonedDateTime.ofInstant(toInstant(), ZonedDateTime.systemDefault()).withZoneSameInstant(ZoneId.of("UTC"))
```

... in which case we may want to rename the method to `toZonedDateTimeUTC`.

app/src/main/java/tech/relaycorp/gateway/domain/courier/GenerateCargo.kt
```diff
@@ -1,11 +1,69 @@
...
+    private val diskMessageOperations: DiskMessageOperations
+) {
 
-    // TODO: implementation
-    suspend fun generate(): Iterable<InputStream> = emptyList()
+    suspend fun generate(): Iterable<() -> InputStream> =
+        (getPCAsMessages() + getParcelMessages())
+            .asSequence()
+            .batch()
+            .map { { it.cargoMessageSet.serializePlaintext().inputStream() } }
```
gnarea, line 31: This is losing `it.latestMessageExpiryDate`, which is required to make the corresponding cargo message as that must be the expiry date of the cargo.

app/src/main/java/tech/relaycorp/gateway/domain/courier/GenerateCargo.kt
```diff
@@ -1,11 +1,69 @@
...
+            .asIterable()
+
+    private suspend fun getPCAsMessages() =
+        parcelCollectionDao.getAll().map { it.toCargoMessageWithExpiry() }
+
+    private suspend fun getParcelMessages() =
+        storedParcelDao.listForRecipientLocation(RecipientLocation.ExternalGateway)
+            .mapNotNull { it.toCargoMessageWithExpiry() }
```
gnarea, line 39: When can it be `null`?

app/src/main/java/tech/relaycorp/gateway/domain/courier/GenerateCargo.kt
```diff
@@ -1,11 +1,69 @@
...
-@Inject constructor() {
+@Inject constructor(
+    private val storedParcelDao: StoredParcelDao,
+    private val parcelCollectionDao: ParcelCollectionDao,
+    private val diskMessageOperations: DiskMessageOperations
+) {
 
-    // TODO: implementation
-    suspend fun generate(): Iterable<InputStream> = emptyList()
+    suspend fun generate(): Iterable<() -> InputStream> =
```
gnarea, line 27: Note that this isn't generating cargoes: It's generating the payload for each cargo, but it's not encrypting the payload nor is it creating the actual cargo itself.

This function should be doing something like:

```
suspend fun generateCargoes(): Sequence<ByteArray> = (generatePCAs() + retrieveParcels())
    .batch()
    .map {
        val cargoMessageSetSerialized = it.cargoMessageSet.encrypt(publicGatewayCert)
        val creationDate = ZonedDateTime.now()
        val ttl = ChronoUnit.SECONDS.between(creationDate, it.latestMessageExpiryDate)
        val cargo = Cargo(
            "https://public-gateway.com",
            privateGatewayCert,
            cargoMessageSetSerialized,
            creationDate = creationDate,
            ttl = ttl
        )
        return cargo.serialize(privateGatewayKeyPair.privateKey)
    }
```

app/src/test/java/tech/relaycorp/gateway/domain/courier/GenerateCargoTest.kt
```diff
@@ -0,0 +1,47 @@
...
+        whenever(storedParcelDao.listForRecipientLocation(any()))
+            .thenReturn(listOf(StoredParcelFactory.build()))
+        whenever(diskMessageOperations.readMessage(any(), any()))
+            .thenReturn("ABC".toByteArray()::inputStream)
+        whenever(parcelCollectionDao.getAll())
+            .thenReturn(listOf(ParcelCollectionFactory.build()))
+
+        val cargo = generateCargo.generate()
+
+        assertEquals(1, cargo.count())
```
gnarea, line 45: I think the most important thing to test here is that the cargo was generated as intended:

- That its `recipientAddress` and `expiryDate` are the correct ones. This would require calling `Cargo.deserialize()`
- That it actually contains the two messages. This would require calling `cargo.unwrapPayload(publicGatewayKeyPair.privateKey)`.

gnarea:
FYI: I've just found out that the test suite isn't run -- I suggest not to merge this until #57 has been fixed so we can identify any regressions more easily.

app/src/main/java/tech/relaycorp/gateway/common/DateUtils.kt
sdsantos: All dates are coming from outside the app so far. And I'm assuming all come in UTC. I store them under a `Date` class, which does not have a timezone, just stores milliseconds. So here I keep assuming there's no conversion going on, it's all still UTC. What we need to be careful is doing a plain `Date()`, cause it will be using the local timezone, and not UTC.

app/src/main/java/tech/relaycorp/gateway/domain/courier/GenerateCargo.kt
sdsantos: When we can't read the Parcel from disk, in the method below. I'll make that method return type explicit, so it's clearer.

app/src/main/java/tech/relaycorp/gateway/domain/courier/GenerateCargo.kt
sdsantos: Just to make sure, the `messageId` goes null? Or do we generate one?

app/src/main/java/tech/relaycorp/gateway/common/DateUtils.kt
sdsantos: Just refactored everything to stop using `Date`s to make our lives easier.

gnarea: Nice, even better!

app/src/main/java/tech/relaycorp/gateway/domain/courier/GenerateCargo.kt
gnarea: It can be unspecified, in which case the core lib will generate a UUID4.

We might want to generate custom message ids if we want to override a cargo message we've already sent though, but we can discuss that later.

gnarea:
LGTM, so happy to approve once the CI issue has been resolved.

PR #56, opened 2020-07-24
fix: Enforce authorisation when processing cargoes and parcels

gnarea:


gnarea:
Let's not merge this until #57 has been fixed.

app/src/sharedTest/java/tech/relaycorp/gateway/test/CargoDeliveryCertPath.kt
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Relaynet PKI certification path from a private gateway to a public gateway.
+ *
+ * See also [FullCertPath]
+ */
+object CargoDeliveryCertPath {
+    private val startDate = ZonedDateTime.now().minusMinutes(1)
+    private val endDate = ZonedDateTime.now().plusHours(1)
+
+    val PRIVATE_GW = issueGatewayCertificate(
```
sdsantos, line 15: I'm afraid all this global `val`s will run whenever a test gets initiated. Can we use instead `val PRIVATE_GW by lazy { issueGatewayCertificate(...) }`? Trying to avoid making the tests slower

gnarea: Absolutely! I didn't realise it'd be re-evaluated.

PR #57, opened 2020-07-24
chore(CI): Actually run unit test suite

gnarea:


app/src/test/java/tech/relaycorp/gateway/ExampleUnitTest.kt
```diff
@@ -4,8 +4,10 @@ import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Test
 
 class ExampleUnitTest {
+    private val very_looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong_variable = 4
+
     @Test
     internal fun test() {
-        assertEquals(4, 2 + 2)
+        assertEquals(very_looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong_variable, 1)
```
gnarea, line 11: This test is broken, but `gradle test` doesn't spot it because it's not running the test suite.

app/src/test/java/tech/relaycorp/gateway/ExampleUnitTest.kt
```diff
@@ -4,8 +4,10 @@ import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Test
 
 class ExampleUnitTest {
+    private val very_looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong_variable = 4
```
gnarea, line 7: This line is longer than 100 chars but `gradle check` doesn't flag it.

gnarea:
I have no idea why this is happening: I couldn't find any obvious difference between the Gradle build files here and those in the courier app :man_shrugging:

PR #65, opened 2020-07-28
fix: Use RSA-PSS when verifying handshake nonces

gnarea:


app/src/main/java/tech/relaycorp/gateway/common/CryptoUtils.kt
```diff
@@ -0,0 +1,7 @@
+package tech.relaycorp.gateway.common
+
+import org.bouncycastle.jce.provider.BouncyCastleProvider
+
+object CryptoUtils {
+    val BC_PROVIDER = BouncyCastleProvider()
```
gnarea, line 6: @sdsantos, are you aware of any potential performance/stability issues with using the security provider this way vs `Security.addProvider()`?

I came across https://bugs.openjdk.java.net/browse/JDK-8168469 but presumably that doesn't apply here because this `BouncyCastleProvider` should only be created once in the lifetime of the process, right?

I wanted to avoid the `Security.addProvider()` approach because I couldn't think of any simple way to call it in the unit/instrumentation test suites, but I'm happy to change tack if you can think of a way to do it.

sdsantos: I'm not familiar with it, but I've had to do something similar for Conscrypt, see `App.setupTLSProvider()`. It basically runs on App `onCreate`, which will also run for Instrumentation tests.

gnarea: Makes sense. I think the tricky bit is making it work in the unit test suite then: I think the only option would be to decorate the tests that directly or indirectly need BC with `@ExtendWith`.

Since we aren't aware of any performance/stability issue, what do you think if we keep the current implementation? We still have to call `setProvider(...)` in the cases where it matters, the difference is whether we pass the string `"BC"` or the `BouncyCastleProvider` instance.

PR #66, opened 2020-07-30
feat: Notify endpoint and Parcel Delivery

sdsantos:


sdsantos:
Double-check the `generateEndpointCertificate` method, cause I'm not sure that's what it should look like.

app/src/main/java/tech/relaycorp/gateway/domain/endpoint/NotifyEndpoint.kt
```diff
@@ -1,11 +1,28 @@
...
-    // TODO: implementation
-    suspend fun notify(endpointAddress: MessageAddress) {}
+    suspend fun notify(endpointAddress: MessageAddress) {
+        endpointDao.get(endpointAddress)
+            ?.let { endpoint ->
+                context.sendBroadcast(
+                    Intent(RelaynetAndroid.ENDPOINT_NOTIFY_ACTION)
+                        .setPackage(endpoint.applicationId),
+                    RelaynetAndroid.GATEWAY_SYNC_PERMISSION
+                )
```
gnarea, line 24: Just wanted to double check: This isn't broadcasting the event to many/all apps on the phone -- Instead, it's limiting the message to the endpoint's app. Right?

I think that's what happening but I wanted to check because of the word "broadcast".

Or is there any chance that third party developers can register intents for other apps? Or simply monitor broadcasts for other apps?

app/src/main/java/tech/relaycorp/gateway/domain/endpoint/RegisterEndpoint.kt
```diff
@@ -1,16 +1,36 @@
...
     suspend fun register(applicationId: String, endpointAddress: MessageAddress) {
         localEndpointDao.insert(LocalEndpoint(endpointAddress, applicationId))
     }
+
+    suspend fun getGatewayCertificate() = localConfig.getCertificate()
+
+    suspend fun generateEndpointCertificate(endpointAddress: MessageAddress) {
+        val gatewayKeyPair = localConfig.getKeyPair()
+        issueEndpointCertificate(
+            gatewayKeyPair.public,
```
gnarea, line 26: This should be the public key of the endpoint, which we need to receive as a parameter instead of the endpoint's address.

app/src/main/java/tech/relaycorp/gateway/domain/endpoint/RegisterEndpoint.kt
```diff
@@ -1,16 +1,36 @@
 package tech.relaycorp.gateway.domain.endpoint
```
gnarea, line 1: I had to issue this cert as part of #72, so I think we can undo the changes to this file (the new method doesn't seem to be used anyway).

app/build.gradle
```diff
@@ -63,6 +63,8 @@ android {
     }
     packagingOptions {
         exclude 'META-INF/*'
+        exclude 'META-INF/licenses/*'
+        exclude '**/attach_hotspot_windows.dll'
```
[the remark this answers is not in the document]
sdsantos: Excluding the `META-INF` ones is pretty common, but the dll isn't. Going to add a comment.

app/src/main/java/tech/relaycorp/gateway/domain/endpoint/NotifyEndpoint.kt
sdsantos: Any app can subscribe to a certain broadcast, but since we're using the `setPackage` filter, only components inside that application will receive them.

PR #67, opened 2020-07-30
feat: Refactor connection observer for internet states

sdsantos:


app/src/main/java/tech/relaycorp/gateway/background/CheckInternetAccess.kt
```diff
@@ -0,0 +1,15 @@
...
+
+class CheckInternetAccess
+@Inject constructor(
+    private val pingRemoteServer: PingRemoteServer
+) {
+    suspend fun check() =
+        pingRemoteServer.pingHostname(INTERNET_TEST_WEBSITE)
+
+    companion object {
+        private const val INTERNET_TEST_WEBSITE = "http://google.com"
```
gnarea, line 13: Can you please use the HTTPS version? Blocked HTTP websites can return `200 OK` if the censor/ISP intercepts the request. They can't do that over TLS.

app/src/main/java/tech/relaycorp/gateway/background/ConnectionState.kt
```diff
@@ -0,0 +1,9 @@
+package tech.relaycorp.gateway.background
+
+sealed class ConnectionState {
+    object Disconnected : ConnectionState()
+    object InternetWithoutRelaynet : ConnectionState()
+    object InternetAndRelaynet : ConnectionState()
```
gnarea, line 6: I'd suggest to replace `Relaynet` with `PublicGateway` for consistency with the rest of the code in the PR and to be more precise.

PR #72, opened 2020-08-06
feat: Complete endpoint registration

### LINKED ISSUE #17: Implement signing of endpoint certificates
Per Relaynet Core (internal PDC requirements).

gnarea:
# Recap

Endpoint registration encompasses two steps:

1. The app should request a _Client Registration Authorization_ (CRA) from the gateway via a bound service. The CRA is what we originally meant to implement as a nonce, but I decided not to call it nonce anymore because I realised it was a lot more than just a nonce.
1. The app should sign the CRA and send it back to the gateway via the PDC server. This request is called _Client Registration Request_ (CRR) and, in addition to the CRA, it also includes the endpoint's public key. When the request is valid, the PDC server issues a new certificate for the endpoint and returns it along with the gateway's certificate -- That response is known as _Client Registration_.

The CRA is valid for 5 seconds. I think should be plenty of time for the app to sign it, wake up the PDC server and `POST` the CRR, but I'd like to hear your thoughts on it.

The CRA is only meant to serve two purposes: Prevent replay attacks and mitigate the risk of an app impersonating another app (if the original CRA grantee leaks the CRA).

CRAs are bound to a specific application id. Theoretically, that means that the same app could reuse it to register multiple endpoints whilst the CRA is still valid. I think that's OK -- I don't see any security threat there.

# Summary of changes

- Rename "nonce" to CRA.
- Implement the generation of CRAs and integrate it in the bound service.
- Implement the processing of CRRs and integrate it in the PDC server. Namely:
  - Refuse invalid CRRs.
  - Register the endpoint by storing its application's id in the DB.
  - Issuing a certificate for the endpoint and return it in the client registration.

Fixes #17.


app/src/androidTest/java/tech/relaycorp/gateway/background/endpoint/EndpointPreRegistrationServiceTest.kt
```diff
@@ -7,50 +7,69 @@ import android.os.Looper
...
-            resultMessage?.what
+            EndpointPreRegistrationService.REGISTRATION_AUTHORIZATION,
+            resultMessage!!.what
         )
-        assertNotNull(resultMessage?.obj)
+
+        // Check we got a valid CRA
+        assertTrue(resultMessage!!.obj is ByteArray)
+        val store = SensitiveStore(InstrumentationRegistry.getInstrumentation().targetContext)
+        val localConfig = LocalConfig(store)
```
gnarea, line 59: I'm bypassing dependency injection -- is there any way to just "ask" for a `localConfig` instance in this test?

app/src/main/java/tech/relaycorp/gateway/background/endpoint/EndpointPreRegistrationService.kt
```diff
@@ -0,0 +1,55 @@
...
+                        PREREGISTRATION_REQUEST -> reply(msg)
+                    }
+                }
+            }
+        )
+        return messenger.binder
+    }
+
+    // TODO: Replace runBlocking with scope.launch
+    internal fun reply(requestMessage: Message) = runBlocking {
```
gnarea, line 39: I was trying to use `scope.launch` but the instrumentation test kept failing. Should I be doing something else here or in the test?

app/src/main/java/tech/relaycorp/gateway/background/endpoint/EndpointPreRegistrationService.kt
```diff
@@ -0,0 +1,55 @@
...
+        // TODO: Dynamically compute the application id
+        val craSerialized = endpointRegistration.authorize("temp-app-id")
```
gnarea, line 41: This is another thing that was breaking in the instrumentation test: I'd like to "discover" the `applicationId` (see: https://stackoverflow.com/q/12918731/129437), but I kept getting the current app's id. I guess that's because of the context in which the test is run, right?

Presumably we'll have to mock this in the test, but I wanted to check I'm not missing something.

app/src/main/java/tech/relaycorp/gateway/background/endpoint/EndpointPreRegistrationService.kt
```diff
@@ -0,0 +1,55 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.runBlocking
+import tech.relaycorp.gateway.background.component
+import tech.relaycorp.gateway.domain.endpoint.EndpointRegistration
+import javax.inject.Inject
+
+class EndpointPreRegistrationService : Service() {
+    private val scope get() = CoroutineScope(Dispatchers.IO)
```
gnarea, line 19: Is this the right scope to use? I wasn't sure whether to add a `SupervisorJob` but I didn't think it'd be necessary here from what I read.

app/src/androidTest/java/tech/relaycorp/gateway/background/endpoint/EndpointPreRegistrationServiceTest.kt
```diff
@@ -7,50 +7,69 @@ import android.os.Looper
...
 
-class NonceRequestServiceTest {
+class EndpointPreRegistrationServiceTest {
 
     @get:Rule
     val serviceRule = ServiceTestRule()
 
     @Test
-    fun requestAndReceiveNonce() {
+    fun requestPreRegistration() = runBlocking {
```
gnarea, line 29: I couldn't get this to work with `run BlockingTest` (and `delay()` instead of `Thread.sleep()`). More on this below.

app/src/androidTest/java/tech/relaycorp/gateway/background/endpoint/EndpointPreRegistrationServiceTest.kt
gnarea: Makes sense, although it seems like Dagger wasn't fully integrated in the instrumentation tests, so I completed that and changed this code in b551c056722c36a8a5947b190230a940df3d8b23. It'd be good to get your feedback there too.

app/src/main/java/tech/relaycorp/gateway/background/endpoint/EndpointPreRegistrationService.kt
gnarea: > In the test, you would need some way to control the scope, either by overriding it, or receiving it through the constructor

How can I do that? I guess I'd have to change something here but I don't know how:

https://github.com/relaycorp/relaynet-gateway-android/blob/b551c056722c36a8a5947b190230a940df3d8b23/app/src/androidTest/java/tech/relaycorp/gateway/background/endpoint/EndpointPreRegistrationServiceTest.kt#L39-L43

> Or you need to make sure the test is runBlocking(Dispatchers.IO) {} I believe

I changed the `runBlocking` to `scope.launch` in the `reply()` and then changed the test to pass `Dispatchers.IO` to `runBlocking`, but now I get the following error:

```
08/11 16:49:14: Launching 'requestPreRegistra...()' on Pixel 2 API 29.
Running tests

$ adb shell am instrument -w -r    -e debug false -e class 'tech.relaycorp.gateway.background.endpoint.EndpointPreRegistrationServiceTest#requestPreRegistration' tech.relaycorp.gateway.test/tech.relaycorp.gateway.test.TestAndroidJUnitRunner
Connected to process 12367 on device 'Pixel_2_API_29 [emulator-5554]'.

Started running tests


Test failed to run to completion. Reason: 'Instrumentation run failed due to 'Process crashed.''. Check device logcat for details
Test running failed: Instrumentation run failed due to 'Process crashed.'
Tests ran to completion.
```

Logcat:

```
2020-08-11 16:52:31.632 12940-12977/tech.relaycorp.gateway E/AndroidRuntime: FATAL EXCEPTION: DefaultDispatcher-worker-1
    Process: tech.relaycorp.gateway, PID: 12940
    java.lang.NullPointerException: Attempt to invoke virtual method 'void android.os.Messenger.send(android.os.Message)' on a null object reference
        at tech.relaycorp.gateway.background.endpoint.EndpointPreRegistrationService$reply$1.invokeSuspend(EndpointPreRegistrationService.kt:43)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:738)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
```

PR #92, opened 2020-08-18
feat: Parcel Collection sending

sdsantos:


app/src/main/java/tech/relaycorp/gateway/pdc/local/PDCServer.kt
```diff
@@ -1,18 +1,54 @@
...
-        registerEndpointRegistration(endpointRegistration)
-        registerParcelCollection()
+    private val server by lazy {
+        embeddedServer(Netty, PORT) {
+            install(WebSockets)
+
+            routing {
+                register(endpointRegistrationRoute)
+                register(parcelCollectionRoute)
+            }
```
gnarea, line 30: From what I've seen, ktor developers encourage a pattern where you have a single function to configure the app (see: [module](https://ktor.io/servers/application.html#modules)), and that function is then called by `embeddedServer` or `withTestApplication`. I think that pattern, or something closer to it, makes sense here too: If we don't do it, we'd have to duplicate all the configuration in this file and then again in the tests (as is the case in the new `testPDCServerRoute` function).

All we're doing right now is to install WebSockets and register a couple of routes, but things are very likely to get more complex: We're likely to do `install()` more features and we're likely to do stuff inside `routing` blocks (like adding `interceptors`). [Interceptors](https://ktor.io/servers/features/routing.html#interception) is something I'm considering using so that we block web browser access across all routes.

I think having an extension function roughly analogous to the original `Application.configure()` would get around this and we'd be able to use it in this property as follows:

```
private val server by lazy {
    embeddedServer(Netty, PORT) {
        configure(arrayOf(endpointRegistrationRoute, parcelCollectionRoute))
    }
}
```

And then in the test suite as follows:

```
fun <R> testPDCServerRoute(route: PDCServerRoute, test: TestApplicationEngine.() -> R) =
    withTestApplication(
        {
            configure(arrayOf(route::register))
        },
        test
    )
```

Thoughts?

sdsantos: Ok, makes sense the approach, specially if we get a lot of configuration commands. The downside is that we will be doing all that configuration always, even if it's no needed for some tests, or we might want to mock those too. But let's see first if it comes to that.

app/src/test/java/tech/relaycorp/gateway/pdc/local/routes/ParcelCollectionRouteTest.kt
```diff
@@ -0,0 +1,114 @@
...
+import tech.relaycorp.gateway.domain.endpoint.CollectParcels
+import tech.relaycorp.gateway.pdc.local.utils.ParcelCollectionHandshake
+import tech.relaycorp.gateway.test.WaitAssertions.suspendWaitFor
+import tech.relaycorp.relaynet.messages.control.ParcelDelivery
+import tech.relaycorp.relaynet.wrappers.x509.Certificate
+import javax.inject.Provider
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+class ParcelCollectionRouteTest {
```
gnarea, line 27: I'd suggest adding tests for:

- Invalid Keep-Alive header value.
- Client closing the TCP connection abruptly. (Assuming we can do it from a unit test; otherwise we can do it in a functional/instrumentation test)
- Client closing the WebSocket connection with normal close code (`1000`)
- Client closing the WS connection with any close code but the normal one.
- Binary ACK messages.

app/src/test/java/tech/relaycorp/gateway/pdc/local/routes/ParcelCollectionRouteTest.kt
```diff
@@ -0,0 +1,114 @@
...
+    fun `Server should keep connection if Keep-Alive is on`() =
+        runBlockingTest {
+            whenever(collectParcels.noParcelsToDeliverOrAck).thenReturn(flowOf(true))
+
+            testPDCServerRoute(route) {
+                handleWebSocketConversation(
+                    ParcelCollectionRoute.URL_PATH,
+                    { addHeader(ParcelCollectionRoute.HEADER_KEEP_ALIVE, "on") }
+                ) { incoming, _ ->
+                    assertTrue(incoming.isEmpty)
```
gnarea, line 78: How is this checking that the connection is kept alive?

This is a tough condition to check I guess. Maybe we need a timeout?

gnarea (changes requested):
It looks fairly easily to follow despite the inherent complexity of the requirements!

app/src/main/java/tech/relaycorp/gateway/domain/endpoint/CollectParcels.kt
```diff
@@ -0,0 +1,76 @@
...
+@Inject constructor(
+    private val storedParcelDao: StoredParcelDao,
+    private val diskMessageOperations: DiskMessageOperations,
+    private val deleteParcel: DeleteParcel
+) {
+
+    private val parcelsSent = mutableListOf<StoredParcel>()
+    private val parcelsWaitingAck = mutableMapOf<String, StoredParcel>()
+
+    private val _noParcelsLeft = MutableStateFlow(false)
```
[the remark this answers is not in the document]
sdsantos: Just my convention for marking the mutable private versions of the state flows. Could be `noParcelsLeftMutable`, or something else.

app/src/main/java/tech/relaycorp/gateway/domain/endpoint/CollectParcels.kt
```diff
@@ -0,0 +1,76 @@
...
+            storedParcel.getInputStream()?.let {
+                Pair(localId, it)
+            }
+        }
+
+    private suspend fun StoredParcel.getInputStream() =
+        try {
+            diskMessageOperations.readMessage(StoredParcel.STORAGE_FOLDER, storagePath)()
+        } catch (e: MessageDataNotFoundException) {
+            logger.log(Level.WARNING, "Read parcel", e)
```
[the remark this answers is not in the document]
sdsantos: It shouldn't happen under normal circumstances, but we never have full control of the user's device. Plus it's an handled error, that's why I kept it as a warning. But I can raise it's level.

app/src/main/java/tech/relaycorp/gateway/pdc/local/routes/ParcelCollectionRoute.kt
```diff
@@ -1,21 +1,39 @@
...
+import tech.relaycorp.relaynet.messages.control.ParcelDelivery
+import tech.relaycorp.relaynet.wrappers.x509.Certificate
 import javax.inject.Inject
+import javax.inject.Provider
 
 class ParcelCollectionRoute
 @Inject constructor(
-    private val parcelCollectionHandshake: ParcelCollectionHandshake
+    private val parcelCollectionHandshake: ParcelCollectionHandshake,
+    private val collectParcelsProvider: Provider<CollectParcels>
```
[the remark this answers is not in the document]
sdsantos: Yup, we want to have one instance per client. But we still want the instances to be provided by dagger. 

It's also useful if you want to delay the object instantiation until you actually need it (lazy).

app/src/main/java/tech/relaycorp/gateway/domain/endpoint/CollectParcels.kt
```diff
@@ -0,0 +1,76 @@
...
+    private val storedParcelDao: StoredParcelDao,
+    private val diskMessageOperations: DiskMessageOperations,
+    private val deleteParcel: DeleteParcel
+) {
+
+    private val parcelsSent = mutableListOf<StoredParcel>()
+    private val parcelsWaitingAck = mutableMapOf<String, StoredParcel>()
+
+    private val _noParcelsLeft = MutableStateFlow(false)
+    val noParcelsToDeliverOrAck: Flow<Boolean> = _noParcelsLeft
```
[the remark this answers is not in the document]
sdsantos: That part is "magical" :P

It's not clear from the logic, because it happens indirectly, through the reactive DB query. When you receive an ACK, the stored parcel gets deleted, so the `storedParcelDao.listForRecipients` returns again.

app/src/main/java/tech/relaycorp/gateway/pdc/local/routes/ParcelCollectionRoute.kt
```diff
@@ -32,13 +50,71 @@ class ParcelCollectionRoute
...
+            // TODO: Call cert.getCertificatePath() and validate certificate chain
 
-            val endpointAddresses =
-                certificates.map { PrivateMessageAddress(it.subjectPrivateAddress) }
+            val collectParcels = collectParcelsProvider.get()
+            sendParcels(collectParcels, certificates.toAddresses())
+            receiveAcks(collectParcels)
+
+            val keepAlive = call.request.header(HEADER_KEEP_ALIVE) == "on"
+            if (!keepAlive) {
```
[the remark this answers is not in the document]
sdsantos: I'll remove the inner check instead, to save our coroutines.

app/src/main/java/tech/relaycorp/gateway/pdc/local/routes/ParcelCollectionRoute.kt
```diff
@@ -28,53 +32,86 @@ class ParcelCollectionRoute
...
+                    if (reason == null) {
+                        logger.log(Level.WARNING, "TCP connection closed abruptly", exc)
+                    } else if (reason.knownReason != CloseReason.Codes.NORMAL) {
+                        logger.log(Level.WARNING, "TCP connection closed due to $reason", exc)
+                    }
+                } catch (exc: Exception) {
+                    logger.log(Level.WARNING, "Connection closed without normal reasons", exc)
+                }
             }
+        } catch (exc: Exception) {
```
gnarea, line 57: Same comment above about catching all exceptions.

app/src/main/java/tech/relaycorp/gateway/pdc/local/routes/ParcelCollectionRoute.kt
```diff
@@ -28,53 +32,86 @@ class ParcelCollectionRoute
...
+                try {
+                    handle()
+                } catch (exc: ClosedReceiveChannelException) {
+                    val reason = closeReason.await()
+                    if (reason == null) {
+                        logger.log(Level.WARNING, "TCP connection closed abruptly", exc)
+                    } else if (reason.knownReason != CloseReason.Codes.NORMAL) {
+                        logger.log(Level.WARNING, "TCP connection closed due to $reason", exc)
+                    }
+                } catch (exc: Exception) {
```
gnarea, line 53: I think it'd be best not to gracefully log and discard any exception, but limit that "graceful" handling to the exception(s) we really want to treat that way. We could be hiding bugs by doing this.

app/src/test/java/tech/relaycorp/gateway/pdc/local/routes/ParcelCollectionRouteTest.kt
```diff
@@ -111,4 +139,57 @@ class ParcelCollectionRouteTest {
...
         }
+
+    @Test
+    fun `Server should ignore binary acks`() =
+        runBlockingTest {
+            testPDCServerRoute(route) {
+                handleWebSocketConversation(ParcelCollectionRoute.URL_PATH) { _, outgoing ->
+                    outgoing.send(Frame.Binary(true, "abc".toByteArray()))
+
+                    verify(collectParcels, never()).processParcelAck(any())
```
gnarea, line 150: I think this may be happening too soon: `collectParcels.processParcelAck()` might actually be called (if there's a bug), it just might not be called when this `verify()` is done. I'd suggest to wait until the connection is closed before doing this assertion.

gnarea (changes requested):

app/src/test/java/tech/relaycorp/gateway/pdc/local/routes/ParcelCollectionRouteTest.kt
```diff
@@ -0,0 +1,195 @@
...
+    fun `Server should handle gracefully client abrupt close`() =
+        runBlockingTest {
+            testPDCServerRoute(route) {
+                handleWebSocketConversation(ParcelCollectionRoute.URL_PATH) { incoming, _ ->
+                    this@testPDCServerRoute.stop(0, 0)
+                    // TODO: figure out a way to tell if the server handled this correctly
+                }
+            }
+        }
+
```
gnarea, line 165: As mentioned on Slack, I think we should abandon this and undo the code to support it in the test suite and the UUT.

app/src/main/java/tech/relaycorp/gateway/pdc/local/routes/ParcelCollectionRoute.kt
sdsantos: This was my un-finished attempt to catch an abrupt client shutdown. Unsuccessful as well...

gnarea (approved):

PR #105, opened 2020-08-28
fix(PDC): Rename private node registration endpoint

gnarea:
Should've done it in #104


gnarea:
Instrumentation test suite failed because the results in one of the devices was "inconclusive", which from what I read online could be an internal issue in Test Lab :man_shrugging: So I'm rerunning the tests.

inconclusive

PR #107, opened 2020-08-28
feat: Ignore duplicated parcels

### LINKED ISSUE #93: Ensure we're not storing and delivering duplicate parcels

sdsantos:
Closes #93

### DIFF
--- a/app/src/main/java/tech/relaycorp/gateway/data/database/ParcelCollectionDao.kt
+++ b/app/src/main/java/tech/relaycorp/gateway/data/database/ParcelCollectionDao.kt
@@ -5,7 +5,10 @@ import androidx.room.Delete
 import androidx.room.Insert
 import androidx.room.OnConflictStrategy
 import androidx.room.Query
+import tech.relaycorp.gateway.data.model.MessageAddress
+import tech.relaycorp.gateway.data.model.MessageId
 import tech.relaycorp.gateway.data.model.ParcelCollection
+import tech.relaycorp.gateway.data.model.PrivateMessageAddress
 
 @Dao
 interface ParcelCollectionDao {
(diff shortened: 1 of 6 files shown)


app/src/main/java/tech/relaycorp/gateway/data/database/ParcelCollectionDao.kt
```diff
@@ -18,4 +21,19 @@ interface ParcelCollectionDao {
...
+            AND senderAddress = :senderAddress 
+            AND messageId = :messageId
+        LIMIT 1
+        """
+    )
+    suspend fun get(
+        recipientAddress: MessageAddress,
+        senderAddress: PrivateMessageAddress,
+        messageId: MessageId
+    ): ParcelCollection?
```
gnarea, line 38: Since we're not interested in the `ParcelCollection` object, but we just want to know if a record exists, I'd suggest to just return a boolean. For example:

```suggestion
    @Query(
        """
        SELECT 1 FROM ParcelCollection
        WHERE recipientAddress = :recipientAddress 
            AND senderAddress = :senderAddress 
            AND messageId = :messageId
        LIMIT 1
        """
    )
    suspend fun exists(
        recipientAddress: MessageAddress,
        senderAddress: PrivateMessageAddress,
        messageId: MessageId
    ): Boolean
```

app/src/main/java/tech/relaycorp/gateway/domain/courier/ProcessCargo.kt
```diff
@@ -49,7 +49,8 @@ class ProcessCargo
         val parcel = when (
             val result = storeParcel.store(parcelData, RecipientLocation.LocalEndpoint)
         ) {
-            is StoreParcel.Result.MalformedParcel -> return
+            is StoreParcel.Result.MalformedParcel,
+            is StoreParcel.Result.DuplicatedParcel -> return
```
gnarea, line 53: I think getting a duplicated parcel shouldn't be logged as a warning that says "Invalid parcel received" because that's misleading. I think it should either not be logged (like in the success case) or logged with a debug/info message that says something like "Ignoring previously-processed parcel".

gnarea (changes requested):

app/src/main/java/tech/relaycorp/gateway/domain/courier/ProcessCargo.kt
sdsantos: You probably misread, cause right now it's not logging anything like the MalformedParcel :)

gnarea: haha, I did! :smile:

gnarea (approved):

PR #121, opened 2020-09-09
chore(testing): Integrate relaynet-testing library

gnarea:


app/build.gradle
```diff
@@ -115,6 +116,8 @@ dependencies {
     implementation 'tech.relaycorp:relaynet:1.36.3'
     implementation 'tech.relaycorp:relaynet-cogrpc:1.1.2'
     implementation 'tech.relaycorp:relaynet-cogrpc-okhttp:1.1.0'
+    testImplementation "tech.relaycorp:relaynet-testing:$relaynetTestingVersion"
+    androidTestImplementation "tech.relaycorp:relaynet-testing:$relaynetTestingVersion"
```
gnarea, line 120: Is there a way to specify this once? I couldn't find one.

sdsantos: Nop, you need to specify it both. But I'm usually relegating the test dependencies to the bottom of the `dependencies` list.

sdsantos (approved):

app/build.gradle
gnarea: Good point! I'll keep in mind in the future for consistency

PR #124, opened 2020-09-10
fix: Pre-registration and open connection

sdsantos:


.gitignore
```diff
@@ -1,6 +1,7 @@
 # Built application files
 *.apk
 *.ap_
+output-metadata.json
```
gnarea, line 4: What's this for?

app/src/main/java/tech/relaycorp/gateway/domain/endpoint/NotifyEndpoint.kt
```diff
@@ -19,8 +19,7 @@ class NotifyEndpoint
             ?.let { endpoint ->
                 context.sendBroadcast(
                     Intent(RelaynetAndroid.ENDPOINT_NOTIFY_ACTION)
-                        .setPackage(endpoint.applicationId),
-                    RelaynetAndroid.GATEWAY_SYNC_PERMISSION
+                        .setPackage(endpoint.applicationId)
```
gnarea, line 22: From a search on GH, it seems like `GATEWAY_SYNC_PERMISSION` isn't used anywhere else. Can you please delete it?

gnarea (changes requested):

.gitignore
sdsantos: When you build an APK, it generates that json to go alongside it.

gnarea (approved):

PR #127, opened 2020-09-14
fix: Improve gateway app based on Demo app experience

sdsantos:


gnarea (commented):
LGTM! It's great to see things generally worked, except for some minor tweaks here and there

app/build.gradle
```diff
@@ -1,6 +1,7 @@
 buildscript {
     repositories {
         maven { url 'https://plugins.gradle.org/m2/' }
+        maven { url 'https://jitpack.io' }
```
gnarea, line 4: I think this was needed at one point but it doesn't seem to be needed anymore.

app/src/main/java/tech/relaycorp/gateway/data/preference/PublicGatewayPreferences.kt
```diff
@@ -30,8 +34,11 @@ class PublicGatewayPreferences
 
     fun getCertificate() = { certificate }.toFlow()
         .map {
-            if (it.isEmpty()) throw RuntimeException("Public Gateway Certificate not set")
-            val certificateBytes = it.toByteArray(Charset.defaultCharset())
+            val certificateBytes = if (it.isEmpty()) {
+                resources.openRawResource(R.raw.public_gateway_cert).readBytesAndClose()
+            } else {
```
gnarea, line 39: FYI: This is OK for now, but we'll probably have to set all the public gateway data upfront, when the app is started for the first time.

gnarea (changes requested):

app/build.gradle
sdsantos: Done

gnarea (approved):

PR #139, opened 2020-09-21
fix: Skip expired parcels

### LINKED ISSUE #128: Cargo generated for couriers may contain expired parcels
The following check uncovered this bug:

https://github.com/relaycorp/relaynet-gateway-android/blob/3850b9d2a0849b01070d0b65700ac40cdbebfce3/app/src/main/java/tech/relaycorp/gateway/domain/courier/GenerateCargo.kt#L89-L93

I think we should change the following:

- Alter the SQL to skip expired parcels and parcel collection acknowledgements. They're purged periodically anyway.
- Replace the exception with a warning. Even if the condition is met due to a future regression, it shouldn't abort the delivery of the cargo. The worst that could happen is that the courier will get an expired cargo, in which case the courier will ACK it but it'll ignore it.

sdsantos:
Closes #128

### DIFF
--- a/app/src/main/java/tech/relaycorp/gateway/data/database/StoredParcelDao.kt
+++ b/app/src/main/java/tech/relaycorp/gateway/data/database/StoredParcelDao.kt
@@ -6,11 +6,13 @@ import androidx.room.Insert
 import androidx.room.OnConflictStrategy
 import androidx.room.Query
 import kotlinx.coroutines.flow.Flow
+import tech.relaycorp.gateway.common.nowInUtc
 import tech.relaycorp.gateway.data.model.MessageAddress
 import tech.relaycorp.gateway.data.model.MessageId
 import tech.relaycorp.gateway.data.model.RecipientLocation
 import tech.relaycorp.gateway.data.model.StorageSize
 import tech.relaycorp.gateway.data.model.StoredParcel
+import java.time.ZonedDateTime
 
 @Dao
 interface StoredParcelDao {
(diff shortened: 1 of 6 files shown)


app/src/main/java/tech/relaycorp/gateway/data/database/StoredParcelDao.kt
```diff
@@ -27,22 +29,28 @@ interface StoredParcelDao {
     @Query(
         """
         SELECT * FROM Parcel
-        WHERE recipientLocation = :recipientLocation 
+        WHERE recipientLocation = :recipientLocation AND expirationTimeUtc > :expiresSince
```
gnarea, line 32: How about we use `date('now')` instead of `:expiresSince`? That parameter isn't currently used by the code calling this method, and it probably shouldn't be set anyway.

According to the SQLite docs, `date('now')` uses UTC: https://www.sqlite.org/lang_datefunc.html

This comment also applies to `listForRecipients`.

gnarea (changes requested):

app/src/main/java/tech/relaycorp/gateway/data/database/StoredParcelDao.kt
sdsantos: We're comparing against epoch milliseconds, so from what I could test on sqlite we would need to compare with something like `datetime(expirationTimeUtc / 1000, 'unixepoch' 'utc') > datetime('now', 'utc')` I believe.

I thought about that. But the fact that we're storing in epoch milliseconds is specified in the kotlin side, on the Converters. So I though I shouldn't be leaking that abstraction to SQL and then locking it there too.

gnarea: I see. I didn't remember SQLite doesn't have proper data types for dates! Let's leave it as it is in that case.

sdsantos:
@gnarea rebased, needs a new review ;)

gnarea (approved):

PR #148, opened 2020-09-29
chore: Parcel Sync instrumented tests

sdsantos:


app/src/main/res/xml/network_security_config.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
```
gnarea, line 1: Can you please add a comment to say we're only doing this so we can connect to the server in the instrumentation tests?

app/src/main/java/tech/relaycorp/gateway/data/disk/SensitiveStore.kt
```diff
@@ -24,6 +24,7 @@ class SensitiveStore
 
     suspend fun store(location: String, data: ByteArray) {
         withContext(Dispatchers.IO) {
+            buildFile(location).delete()
```
gnarea, line 27: Does this work when the file doesn't exist yet? Can you please write a test for this?

app/build.gradle
```diff
@@ -179,6 +179,7 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
     kotlinOptions.freeCompilerArgs += [
             '-Xuse-experimental=kotlinx.coroutines.ExperimentalCoroutinesApi',
             '-Xuse-experimental=kotlinx.coroutines.FlowPreview',
-            '-Xuse-experimental=kotlin.time.ExperimentalTime'
+            '-Xuse-experimental=kotlin.time.ExperimentalTime',
+            '-Xuse-experimental=io.ktor.util.KtorExperimentalAPI'
```
gnarea, line 183: Is there anything we can do in the PoWeb library to prevent devs from having to do this? I couldn't find the answer online

gnarea (changes requested):

app/build.gradle
sdsantos: I'm not sure, because part of Ktor is exposed to us. But I can look into it.

gnarea: Thanks. I'd imagine this is the culprit: https://github.com/relaycorp/relaynet-poweb-jvm/blob/0fcb768994cc6b9f287e371bdfd9c617041b8330/src/main/kotlin/tech/relaycorp/poweb/PoWebClient.kt#L53

But I don't know how to prevent it from propagating to clients of the PoWeb lib.

sdsantos: @gnarea wouldn't the same compilerArgs approach I did work for the library? But I will look this up.

gnarea: Thanks for looking into this! Can you please use the new version of PoWeb so we can use the fix in https://github.com/relaycorp/relaynet-poweb-jvm/pull/62 and undo this workaround?

gnarea:
@sdsantos, following up on https://github.com/relaycorp/relaynet-gateway-android/issues/149#issuecomment-701306975, I think `parcelDelivery_invalidParcel` is now failing because the server isn't available when we make the request:

```
java.lang.Exception: Unexpected exception, expected<tech.relaycorp.poweb.RejectedParcelException> but was<tech.relaycorp.poweb.ServerConnectionException>
(...)
Caused by tech.relaycorp.poweb.ServerConnectionException: Failed to connect to http://127.0.0.1:13276/v1/parcels
(...)
Caused by: java.net.ConnectException: Failed to connect to /127.0.0.1:13276
(...)
Caused by: java.net.ConnectException: failed to connect to /127.0.0.1 (port 13276) from /127.0.0.1 (port 43068) after 10000ms: (...)
Caused by: android.system.ErrnoException: isConnected failed: ECONNREFUSED (Connection refused)
```

gnarea:
Wait what!? My last two commits seem to have fixed the instrumentation test suite. But I'd just tested them in a separate branch where I implemented them as a POC and the test failed with the error above: https://github.com/relaycorp/relaynet-gateway-android/compare/parcel-delivery-instrumented-test...IOException-bugfix-poc

I suspect this is still an intermittent issue, but maybe my commits made them less frequent :man_shrugging: I'll re-run the CI for the last commit.

gnarea:
Yeah, my latest changes have unintentionally made the issue less frequent, but the test is still flaky: https://console.firebase.google.com/u/0/project/relaynet-gateway-android/testlab/histories/bh.fce322cdd2254629/matrices/6458808712135122226 (that's for the last CI run)

I think the test should wait until the server is ready to receive connections.

app/src/androidTest/java/tech/relaycorp/gateway/background/endpoint/GatewaySyncServiceParcelCollectionTest.kt
```diff
@@ -0,0 +1,107 @@
...
+                        )
+                    ),
+                    StreamingMode.CloseUponCompletion
+                )
+                .first()
+
+        assertEquals(
+            Parcel.deserialize(parcel).id,
+            Parcel.deserialize(parcelCollection.parcelSerialized).id
+        )
```
gnarea, line 101: This is never executed, right? I'd suggest to delete it along with the `parcelCollection` variable so it's clear that `PoWebClient.initLocal()...` should throw the exception

gnarea (changes requested):

gnarea (approved):

PR #157, opened 2020-10-12
feat: Settings About

sdsantos:


app/src/main/java/tech/relaycorp/gateway/ui/settings/LicensesActivity.kt
```diff
@@ -0,0 +1,13 @@
+package tech.relaycorp.gateway.ui.settings
+
+import android.os.Bundle
+import com.mikepenz.aboutlibraries.ui.LibsActivity
+import tech.relaycorp.gateway.R
+
+class LicensesActivity : LibsActivity() {
+    override fun onCreate(savedInstanceState: Bundle?) {
```
gnarea, line 8: Can we limit the dependencies to runtime ones, excluding test and built-time dependencies?

gnarea (changes requested):
There's a tinting issue on Android 5:

Screenshot_2020-10-12-14-51-20
Screenshot_2020-10-12-14-51-33

app/src/main/java/tech/relaycorp/gateway/ui/settings/LicensesActivity.kt
sdsantos: Test dependencies only show up in `debug` builds, according to the documentation.

gnarea: Nice!

gnarea (dismissed):

PR #207, opened 2020-11-27
feat: Public Sync error handling

sdsantos:


app/src/main/java/tech/relaycorp/gateway/domain/publicsync/CollectParcelsFromGateway.kt
```diff
@@ -37,17 +41,31 @@ class CollectParcelsFromGateway
             poWebClient.use {
                 poWebClient
                     .collectParcels(arrayOf(signer), streamingMode)
+                    .retry(Long.MAX_VALUE) { e ->
+                        if (keepAlive && e is ServerException) {
```
gnarea, line 45: I think we shouldn't retry if the failure is due to a protocol violation (`ServerBindingException`): Retrying in a few seconds won't change anything. For that reason, I'd suggest to retry if the exception is a `ServerConnectionException`.

Now, this raises the question of how to handle protocol violations. In an ideal world, I think we should notify the user about it. However, since we'll be operating the initial server, I think that's very unlikely so we should just log that event.

app/src/main/java/tech/relaycorp/gateway/domain/publicsync/CollectParcelsFromGateway.kt
```diff
@@ -37,17 +41,31 @@ class CollectParcelsFromGateway
...
+                                "Could not collect parcels due to server error, will retry.",
+                                e
+                            )
+                            delay(RETRY_AFTER_PERIOD)
+                            true
+                        } else false
+                    }
                     .collect { collectParcel(it, keepAlive) }
             }
         } catch (e: ServerException) {
```
gnarea, line 57: Same comment as above (sorry, I missed it before)

app/src/main/java/tech/relaycorp/gateway/domain/publicsync/CollectParcelsFromGateway.kt
```diff
@@ -37,17 +41,31 @@ class CollectParcelsFromGateway
...
-            logger.log(Level.INFO, "Could not collect parcels due to server error", e)
+            logger.log(Level.SEVERE, "Could not collect parcels due to server error", e)
             return
         } catch (e: ClientBindingException) {
             logger.log(Level.SEVERE, "Could not collect parcels due to client error", e)
             return
         } catch (e: NonceSignerException) {
             logger.log(Level.SEVERE, "Could not collect parcels due to signing error", e)
             return
+        } catch (e: Exception) {
```
gnarea, line 66: I'd suggest to use `PDCException` instead, so that we limit this to PoWeb exceptions without silently masking app bugs.

app/src/main/java/tech/relaycorp/gateway/domain/publicsync/DeliverParcelsToGateway.kt
```diff
@@ -54,7 +54,9 @@ class DeliverParcelsToGateway
                 try {
                     deliverParcel(poWebClient, parcel)
                 } catch (e: ServerException) {
-                    logger.log(Level.INFO, "Could not deliver parcels due to server error", e)
+                    logger.log(Level.INFO, "Could not deliver parcel due to server error", e)
+                } catch (e: Exception) {
```
gnarea, line 58: Same comments as above, re: parcel collections (and apologies that I missed this catch for `ServerException` before!).

Now that I'm looking at this, when is the parcel deleted after delivery? In addition to deleting the parcel upon successful delivery, I think we should also delete when we get a `ForbiddenException`: That means that the Android Gateway isn't authorised to deliver parcels to that public gateway. And even if the Android GW re-registered, the parcel would still be using the old certificate chain, so the parcel could never be delivered.

Finally, can you please add tests for these changes?

gnarea (changes requested):

app/src/main/java/tech/relaycorp/gateway/domain/publicsync/DeliverParcelsToGateway.kt
sdsantos: Forbidden exception is handled by the library and thrown as a RejectedParcelException. And we're already handling that, deleting the parcel and testing it.

gnarea: Good point re: deleting the parcel! In that case, my only suggestion here is to use `PDCException` instead of `Exception`

gnarea (approved):
lgtm :+1:

PR #215, opened 2020-11-30
fix: Remote pings leak

### LINKED ISSUE #184: Internet and Public Gateway pings are leaking connections
After a while with the Gateway app opened, warnings start appearing stating:

```
OkHttpClient: A connection to https://poweb-test.relaycorp.tech/ was leaked. Did you forget to close a response body?
OkHttpClient: A connection to https://google.com/ was leaked. Did you forget to close a response body?
```

We need to fix the code making the ping calls so connections are properly closed.

sdsantos:
Closes #184

### DIFF
--- a/app/src/main/java/tech/relaycorp/gateway/background/PingRemoteServer.kt
+++ b/app/src/main/java/tech/relaycorp/gateway/background/PingRemoteServer.kt
@@ -2,10 +2,12 @@ package tech.relaycorp.gateway.background
 
 import io.ktor.client.HttpClient
 import io.ktor.client.engine.android.Android
+import io.ktor.client.features.UserAgent
 import io.ktor.client.request.head
 import io.ktor.network.selector.ActorSelectorManager
 import io.ktor.network.sockets.aSocket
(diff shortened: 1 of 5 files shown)


app/src/main/java/tech/relaycorp/gateway/background/PingRemoteServer.kt
```diff
@@ -1,55 +1,40 @@
...
-                false
-            } catch (ex: Exception) {
-                logger.log(Level.INFO, "Could not reach $address:$port", ex)
-                false
-            } finally {
-                socket.close()
-            }
+
+    private val ktorClient by lazy {
+        HttpClient(Android)
```
gnarea, line 18: Can you please add the user agent? `Relaynet Private Gateway/<version> (Android)`. Or it isn't easy to get the version of the app, just `Relaynet Private Gateway (Android)`.

That'd be helpful when I'm looking at the server logs.

gnarea (changes requested):

PR #217, opened 2020-11-30
fix: CCA with creation time 5 minutes before

### LINKED ISSUE #204: CCA generation doesn't account for clock drift
So if the courier app is even a second behind the android gw, the CCA will be refused.

We avoid this issue when generating cargo by setting the cargo's creation time to 5 mins ago.

Note we need to be careful not to introduce the same bug in https://github.com/relaycorp/relaynet-gateway-android/issues/205 when we fix this issue.

sdsantos:
Closes #204

### DIFF
--- a/app/src/main/java/tech/relaycorp/gateway/domain/courier/GenerateCCA.kt
+++ b/app/src/main/java/tech/relaycorp/gateway/domain/courier/GenerateCCA.kt
@@ -1,8 +1,10 @@
 package tech.relaycorp.gateway.domain.courier
 
+import tech.relaycorp.gateway.common.nowInUtc
 import tech.relaycorp.gateway.data.preference.PublicGatewayPreferences
 import tech.relaycorp.gateway.domain.LocalConfig
 import tech.relaycorp.relaynet.messages.CargoCollectionAuthorization
+import java.util.Collections.max
 import javax.inject.Inject
 import kotlin.time.days
 
@@ -17,11 +19,20 @@ class GenerateCCA
             recipientAddress = publicGatewayPreferences.getCogRPCAddress(),
             payload = "".toByteArray(),
             senderCertificate = localConfig.getCertificate(),
+            creationDate = calculateCreationDate(),
             ttl = TTL.inSeconds.toInt()
         )
 
     suspend fun generateByteArray() =
         generate().serialize(localConfig.getKeyPair().private)
+
+    private suspend fun calculateCreationDate() =
+        max(
+            listOf(
+                nowInUtc().minusMinutes(5), // Allow for some clock-drift
+                localConfig.getCertificate().startDate // Never before the GW registration
+            )
+        )
 
     companion object {
         private val TTL = 14.days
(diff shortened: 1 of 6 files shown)


gnarea (changes requested):
This code duplicates #216 so I'd suggest to have a single implementation of the `calculateCreationDate()` function (e.g., `calculateCRCMessageCreationDate()`).

A CRC (Cargo Relay Connection) message is any message exchanged between gateways via couriers. Right now there's only CCAs and cargoes, but there could be more in the future.

app/src/main/java/tech/relaycorp/gateway/domain/courier/GenerateCCA.kt
```diff
@@ -9,14 +9,16 @@ import kotlin.time.days
...
+    private val calculateCreationDate: CalculateCRCMessageCreationDate
 ) {
 
     suspend fun generate() =
         CargoCollectionAuthorization(
             recipientAddress = publicGatewayPreferences.getCogRPCAddress(),
             payload = "".toByteArray(),
             senderCertificate = localConfig.getCertificate(),
+            creationDate = calculateCreationDate.calculate(),
             ttl = TTL.inSeconds.toInt()
```
gnarea, line 22: I missed this in the first version of this PR: The TTL should be relative to the `creationDate`, which is going to be particularly important with this PR since the creation date could be the start date of the cert.

We solved this in the cargo generation by using `Duration.between()`, so we should probably factor that out like you've done with `CalculateCRCMessageCreationDate` -- perhaps re-purposing that class to add a second method like `calculateTTL()`.

gnarea (changes requested):

app/src/main/java/tech/relaycorp/gateway/domain/courier/GenerateCCA.kt
sdsantos: But it would be 5 minutes ago or sooner, while the TTL is 14 days. It won't matter much would it?

gnarea: You're right that with the current version of the code it won't make any difference. However, I'm planning to make a change at some point before the stable release that could make a difference:

I'm planning to change all the dates on cargoes and CCAs so that they're totally inaccurate, and this way an eavesdropper couldn't infer at what time the user and the courier met. The meeting time could be used to infer location. For this reason, the plan is to set the creation date to a random number of days in the past. I haven't yet figured out how big to make this range.

sdsantos: Then the TTL should also have some randomness to it I imagine. But I can make the TTL calculated for now as well.

gnarea: Good point! Yeah, the TTL should be the sum of the random duration in the past + the actual TTL. So this future change will be more elaborate than I thought anyway, so let's leave this as it is.

gnarea (approved):

PR #235, opened 2020-12-29
fix(CogRPC): Use unified version of CogRPC address

gnarea:
This change depends on https://github.com/relaycorp/relaynet-courier-android/pull/264

### DIFF
--- a/app/src/main/java/tech/relaycorp/gateway/data/preference/PublicGatewayPreferences.kt
+++ b/app/src/main/java/tech/relaycorp/gateway/data/preference/PublicGatewayPreferences.kt
@@ -65,6 +65,6 @@ class PublicGatewayPreferences
     companion object {
         @VisibleForTesting
         internal const val DEFAULT_ADDRESS = "https://poweb-frankfurt.relaycorp.cloud"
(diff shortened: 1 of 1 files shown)


app/src/main/java/tech/relaycorp/gateway/data/preference/PublicGatewayPreferences.kt
```diff
@@ -65,6 +65,6 @@ class PublicGatewayPreferences
     companion object {
         @VisibleForTesting
         internal const val DEFAULT_ADDRESS = "https://poweb-frankfurt.relaycorp.cloud"
-        private const val COGRPC_ADDRESS = "https://cogrpc-frankfurt.relaycorp.cloud"
+        private const val COGRPC_ADDRESS = "https://frankfurt.relaycorp.cloud"
```
gnarea, line 68: `DEFAULT_ADDRESS` will become `"https://frankfurt.relaycorp.cloud"` in a future PR too, so when we get there I'll delete `COGRPC_ADDRESS`.

sdsantos (approved):

PR #236, opened 2020-12-30
fix(PoWeb): Look up SRV record to resolve PoWeb address

gnarea:
Also remove CogRPC-specific address


app/src/main/java/tech/relaycorp/gateway/data/DataModule.kt
```diff
@@ -74,15 +75,19 @@ class DataModule {
 
     // PoWeb
 
+    @Provides
+    fun dohClient() =
+        // TODO: Remove custom DNS resolver once we can use CloudFlare's (which is the default)
+        // https://github.com/cloudflare/cloudflare-docs/issues/565
+        DoHClient("https://dns.google/dns-query")
```
gnarea, line 82: Since I'm not super familiar with Dagger and Android, I should probably highlight what I intend to do here in case I'm missing something:

I want to keep one DoH client around, so that the same instance can be reused when making queries. In the current version of the code, only the PoWeb client and the URL ping use this DoH client. This is to be able to reuse the underlying resources (e.g., connection pool, keep-alive TCP connections) and [eventually caching](https://github.com/relaycorp/doh-jvm/issues/8).

app/src/main/java/tech/relaycorp/gateway/data/preference/PublicGatewayPreferences.kt
```diff
@@ -24,11 +27,34 @@ class PublicGatewayPreferences
...
-    suspend fun getAddress() = observeAddress().first()
-    private fun observeAddress() = { address }.toFlow()
+    private fun getAddress(): String = address.get()
     suspend fun setAddress(value: String) = address.setAndCommit(value)
 
-    suspend fun getCogRPCAddress() = COGRPC_ADDRESS
+    fun getCogRPCAddress() = "https://$DEFAULT_ADDRESS"
+
+    @Throws(PublicAddressResolutionException::class)
+    suspend fun resolvePoWebAddress(): ServiceAddress {
```
gnarea, line 36: Most of the code in this method is a slight variable of https://github.com/relaycorp/relaynet-courier-android/blob/e5ebacfb91f8590aef0c7726a49dd1719ae572e9/app/src/main/java/tech/relaycorp/courier/data/model/MessageAddress.kt#L44-L61

I think we'll want to move this logic to a shared library in the future, but I didn't want to add it to the core library for two reasons:

- The dependencies of the DoH client are relatively "heavy" and this whole logic isn't needed on the Relaynet-compatible apps.
- DoH is just an implementation detail and Relaynet itself doesn't actually _require_ it. Relaynet only requires DNSSEC on the "unified address" (e.g., `frankfurt.relaycorp.cloud`), and I'm simply using DoH to get DNSSEC working in a quick and safe way.

app/src/main/java/tech/relaycorp/gateway/data/DataModule.kt
sdsantos: To keep it around, just add a dagger Scope to it, and it will be shared inside the scope. If it's the `@Singleton` scope, it will be shared across the whole app.

app/src/main/java/tech/relaycorp/gateway/data/preference/PublicGatewayPreferences.kt
sdsantos: Yeah, they are very similar, we could extract that to a resolver that both classes use. That way it would be easier to mock as well, and test separately.

On the MessageAddress is not as easy to mock, since it's a data model, and not something we inject. But on the resolver extracted class, we could have a global method that does the same thing, to simplify that change.

gnarea: Agreed! I've just created https://github.com/relaycorp/relayverse/issues/20 to track that work

app/src/main/java/tech/relaycorp/gateway/data/DataModule.kt
gnarea: Thanks! I think `@Singleton` will do, having read the docs. I've just added it.

sdsantos (approved):

PR #239, opened 2020-12-31
feat: Public Gateway Migration

sdsantos:


gnarea (changes requested):
You mentioned this PR was WIP, but I just want check that the final version will do the migration in a separate screen like in https://www.figma.com/file/vz5uB0KQyH2up8paufV1LY/Relaynet-%F0%9F%93%A6?node-id=854%3A288 ? I wasn't sure if this was part of the UI improvements you mentioned last week.

Also, I'm setting up a temporary gateway to test this. It should be available on `gwmigrationtest.relaycorp.cloud` in a few minutes. I'll delete it when we merge this.

gnarea:
FYI: I was testing something unrelated to this PR, but I had left this branch installed accidentally, and I discovered that CCAs are still generated with the old gateway address. I added a log statement to confirm this.

This happened after successfully pairing with `gwmigrationtest.relaycorp.cloud`.

app/src/main/java/tech/relaycorp/gateway/data/DataModule.kt
```diff
@@ -83,12 +94,20 @@ class DataModule {
...
-    fun poWebClientBuilder(publicGatewayPreferences: PublicGatewayPreferences) =
-        object : PoWebClientBuilder {
-            @Throws(PublicAddressResolutionException::class)
-            override suspend fun build(): PoWebClient {
-                val address = publicGatewayPreferences.resolvePoWebAddress()
-                return PoWebClient.initRemote(address.host, address.port)
-            }
-        }
+    fun poWebClientBuilder() = object : PoWebClientBuilder {
+        @Throws(PublicAddressResolutionException::class)
```
gnarea, line 98: This `@Throws` is no longer applicable with this change.

app/src/main/java/tech/relaycorp/gateway/domain/publicsync/MigrateGateway.kt
```diff
@@ -1,43 +1,30 @@
...
-        return when (registerGateway.registerIfNeeded()) {
-            is RegisterGateway.Result.Registered,
-            RegisterGateway.Result.AlreadyRegistered -> Result.Successful
+    suspend fun migrate(address: String) =
+        when (registerGateway.registerNewAddress(address)) {
+            is RegisterGateway.Result.Registered -> Result.Successful
             RegisterGateway.Result.FailedToRegister -> Result.FailedToRegister
             RegisterGateway.Result.FailedToResolve -> Result.FailedToResolve
+            RegisterGateway.Result.AlreadyRegistered -> {
+                deleteInvalidatedData()
```
gnarea, line 20: I don't think `AlreadyRegistered` can be returned by `registerNewAddress()`, but if that were the case, I don't think we should delete anything, right?

app/src/main/java/tech/relaycorp/gateway/data/database/StoredParcelDao.kt
```diff
@@ -24,6 +24,9 @@ interface StoredParcelDao {
     @Delete
     suspend fun delete(message: StoredParcel)
 
+    @Query("DELETE FROM ParcelCollection")
+    suspend fun deleteAll()
```
gnarea, line 28: This isn't used, and it refers to the wrong table.

gnarea (changes requested):

app/src/main/java/tech/relaycorp/gateway/domain/publicsync/MigrateGateway.kt
sdsantos: I mixed the 2 Results up, this is the Registered one.

gnarea:
BTW, the new copy seems to break the UI because it pushes down the text field and the interface isn't scrollable, so the user can't see the tooltip or press the submit button.

Screenshot_2021-01-11-12-32-48

gnarea (approved):

PR #250, opened 2021-01-08
chore: Write up README

### LINKED ISSUE #1: Write up README

gnarea:
Fixes #1

I considered describing the architecture and tech stack a bit more, like we did in the Android Courier, but I couldn't think of anything that wasn't obvious to an experienced Android developer (e.g., supported Gradle tasks) or very easy to figure out (we use Kotlin instead of Java). Did I miss something important?


README.md
```diff
@@ -1,7 +1,64 @@
...
+
+The local communication with endpoints does not use TLS, but all other connections are external and therefore require TLS.
+
+By default, instances of this gateway are paired to [Relaycorp's Frankfurt gateway](https://github.com/relaycorp/cloud-gateway/tree/main/environments/frankfurt).
+
+## Security and privacy considerations
+
+The items below summarize the security and privacy considerations specific to this app. For a more general overview of the security considerations in Relaynet, please refer to [RS-019](https://specs.relaynet.network/RS-019).
+
+### No encryption at rest on Android 5
```
sdsantos, line 25: I was not familiar with the expression "at rest" in this context, but it looks to be a known term in computer science.

sdsantos (dismissed):
The README looks good. The only things I would add is what I wrote in the Courier App, architecture and gradle tasks, which you already mentioned. If you believe they aren't important to add, it's fine as it is.

gnarea:
Thanks. In hindsight, I think the architecture bit (https://github.com/relaycorp/relaynet-courier-android#architecture) would be relevant, but I believe things are slightly different on this app, so if you could add that section that'd be great. I'm still not sure about the Gradle tasks because they're all standard (except for `spotless(Check|Apply)`).

sdsantos (approved):

Issue #266, opened 2021-01-15
Self-issued certificate isn't re-generated before expiry


gnarea:
So courier sync will stop working when the certificate expires (it's current validity period is one year).

https://github.com/relaycorp/relaynet-gateway-android/blob/ec06dfc429e4db8e4cde02942c772486991c6898/app/src/main/java/tech/relaycorp/gateway/domain/LocalConfig.kt#L82-L91

sdsantos:
@gnarea Just checked the current logic we have in place.

- We create a new identity certificate the first time the user uses the app
- Expired certificates are deleted automatically on every app start
- If we don't have an identity certificate, we generate and store a new one

So I feel we don't have this issue anymore.

Additionally, though I'm not entirely sure this is related:
- We store an identity certificate after registering with the public gateway
- If we get a CERTIFICATE_ROTATION cargo message from the Courier, we store the new identity certificate that comes inside it, if it expires later than the current one we have

gnarea:
Thanks for looking into this @sdsantos!

The private gateway should have two certificates (for the same identity key pair):

- The one issued by its Internet gateway, which is used when the Internet is available (e.g., parcel collection via PoWeb).
- A self-issued one for the comms via couriers (cargo delivery and collection via CogRPC).

I think we're only talking about the latter here, in which case I agree this isn't an issue anymore if we're purging expired certificates and (re-)creating them on demand when syncing with couriers, so I'll close this issue.

PR #267, opened 2021-01-15
fix: Self-issue certificate when app is used for the first time

gnarea:
The problem here is that the certificate's start date was always later than the CCA's or the cargo's (whose creation time is 90 mins in the past).

I'm proposing to fix this by creating this certificate upfront, along with its key pair.

Fixes https://github.com/relaycorp/relaynet-courier-android/issues/277

### DIFF
--- a/app/src/androidTest/java/tech/relaycorp/gateway/background/endpoint/EndpointPreRegistrationServiceTest.kt
+++ b/app/src/androidTest/java/tech/relaycorp/gateway/background/endpoint/EndpointPreRegistrationServiceTest.kt
@@ -43,7 +43,7 @@ class EndpointPreRegistrationServiceTest {
     fun setUp() {
         AppTestProvider.component.inject(this)
         runBlocking {
-            localConfig.generateKeyPair()
+            localConfig.bootstrap()
             publicGatewayPreferences.setRegistrationState(RegistrationState.Done)
         }
     }
--- a/app/src/androidTest/java/tech/relaycorp/gateway/background/endpoint/EndpointPreRegistrationServiceTest.kt
+++ b/app/src/androidTest/java/tech/relaycorp/gateway/background/endpoint/EndpointPreRegistrationServiceTest.kt
@@ -45,6 +45,7 @@ class EndpointPreRegistrationServiceTest {
     fun setUp() {
         AppTestProvider.component.inject(this)
         runBlocking(coroutineContext) {
+            localConfig.bootstrap()
             publicGatewayPreferences.setRegistrationState(RegistrationState.Done)
         }
     }
--- a/app/src/androidTest/java/tech/relaycorp/gateway/background/endpoint/EndpointPreRegistrationServiceTest.kt
+++ b/app/src/androidTest/java/tech/relaycorp/gateway/background/endpoint/EndpointPreRegistrationServiceTest.kt
@@ -21,6 +21,7 @@ import tech.relaycorp.gateway.data.model.RegistrationState
 import tech.relaycorp.gateway.data.preference.PublicGatewayPreferences
 import tech.relaycorp.gateway.domain.LocalConfig
 import tech.relaycorp.gateway.test.AppTestProvider
+import tech.relaycorp.gateway.test.WaitAssertions.suspendWaitFor
 import tech.relaycorp.gateway.test.WaitAssertions.waitFor
 import tech.relaycorp.relaynet.messages.control.PrivateNodeRegistrationAuthorization
 import java.nio.charset.Charset
@@ -43,7 +44,7 @@ class EndpointPreRegistrationServiceTest {
     fun setUp() {
         AppTestProvider.component.inject(this)
         runBlocking {
-            localConfig.bootstrap()
+            suspendWaitFor { localConfig.getKeyPair() }
             publicGatewayPreferences.setRegistrationState(RegistrationState.Done)
         }
     }
--- a/app/src/androidTest/java/tech/relaycorp/gateway/background/endpoint/EndpointPreRegistrationServiceTest.kt
+++ b/app/src/androidTest/java/tech/relaycorp/gateway/background/endpoint/EndpointPreRegistrationServiceTest.kt
@@ -20,6 +20,7 @@ import tech.relaycorp.gateway.data.model.RegistrationState
 import tech.relaycorp.gateway.data.preference.PublicGatewayPreferences
 import tech.relaycorp.gateway.domain.LocalConfig
 import tech.relaycorp.gateway.test.AppTestProvider
+import tech.relaycorp.gateway.test.WaitAssertions.suspendWaitFor
 import tech.relaycorp.gateway.test.WaitAssertions.waitFor
 import tech.relaycorp.relaynet.messages.control.PrivateNodeRegistrationAuthorization
 import java.nio.charset.Charset
@@ -45,7 +46,7 @@ class EndpointPreRegistrationServiceTest {
     fun setUp() {
         AppTestProvider.component.inject(this)
         runBlocking(coroutineContext) {
-            localConfig.bootstrap()
+            suspendWaitFor { localConfig.getKeyPair() }
             publicGatewayPreferences.setRegistrationState(RegistrationState.Done)
         }
     }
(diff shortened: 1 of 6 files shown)


app/src/main/java/tech/relaycorp/gateway/App.kt
```diff
@@ -117,9 +121,11 @@ open class App : Application() {
         Security.insertProviderAt(Conscrypt.newProvider(), 1)
     }
 
-    private fun registerGateway() {
-        if (mode == Mode.Test) return
+    private fun bootstrapGateway() {
         ioScope.launch {
+            localConfig.generateKeyPair()
+            localConfig.generateCargoDeliveryAuth()
```
gnarea, line 127: I originally implemented this in a new method inside `App`, but I was running into what appears to be a race condition because `registerGateway` and my new method were executed in different threads. This caused the key pair to be generated twice, often causing the cargo delivery auth certificate to be issued with the old key pair.

So I'm repurposing this method so that I can use the same scope.

app/src/main/java/tech/relaycorp/gateway/domain/LocalConfig.kt
```diff
@@ -25,8 +27,9 @@ class LocalConfig
         sensitiveStore.read(PRIVATE_KEY_FILE_NAME)
             ?.toPrivateKey()
             ?.toKeyPair()
-            ?: generateKeyPair()
-                .also { setKeyPair(it) }
+            ?: throw RuntimeException("No key pair was found")
```
gnarea, line 30: The issue I described earlier, where a second key pair was generated and saved, made me realise that we shouldn't be gracefully doing this. This key pair should only be generated once in the lifetime of the app.

Having said this, I wasn't sure how to fail when this method was called before the key pair was available, so a `RuntimeException` might not be a good option.

app/src/main/java/tech/relaycorp/gateway/domain/LocalConfig.kt
```diff
@@ -51,8 +54,10 @@ class LocalConfig
     suspend fun getCargoDeliveryAuth() =
         sensitiveStore.read(CDA_CERTIFICATE_FILE_NAME)
             ?.let { Certificate.deserialize(it) }
-            ?: generateCertificate()
-                .also { setCargoDeliveryAuth(it) }
+            ?: throw RuntimeException("No CDA was found")
```
gnarea, line 57: ditto

app/src/main/java/tech/relaycorp/gateway/domain/LocalConfig.kt
```diff
@@ -74,13 +79,13 @@ class LocalConfig
...
-    private fun generateKeyPair() = generateRSAKeyPair()
-
     private suspend fun generateCertificate(): Certificate {
         val keyPair = getKeyPair()
         return issueGatewayCertificate(
             subjectPublicKey = keyPair.public,
             issuerPrivateKey = keyPair.private,
+            validityStartDate = nowInUtc()
+                .minus(CalculateCRCMessageCreationDate.CLOCK_DRIFT_TOLERANCE.toJavaDuration()),
             validityEndDate = nowInUtc().plusYears(1)
```
gnarea, line 89: FYI: https://github.com/relaycorp/relaynet-gateway-android/issues/266

app/src/main/java/tech/relaycorp/gateway/domain/LocalConfig.kt
```diff
@@ -51,8 +54,10 @@ class LocalConfig
     suspend fun getCargoDeliveryAuth() =
         sensitiveStore.read(CDA_CERTIFICATE_FILE_NAME)
             ?.let { Certificate.deserialize(it) }
-            ?: generateCertificate()
-                .also { setCargoDeliveryAuth(it) }
+            ?: throw RuntimeException("No CDA issuer was found")
```
gnarea, line 57: ditto re: `RuntimeException`

app/src/main/java/tech/relaycorp/gateway/App.kt
sdsantos: Right now this is being called on every app start, and not only as needed. So you're going to get new keypairs every time. The methods should check themselves if they are needed, or we should check before calling them.

Also, they won't run in instrumented tests, which I'm not sure that's what we want. We don't want to run the registration, because that requires other apps in the phone, but generating the keypair I think we want to.

app/src/main/java/tech/relaycorp/gateway/domain/LocalConfig.kt
sdsantos: RuntimeException is fine, it's pretty clear it's not something we want to handle.

sdsantos (changes requested):

app/src/main/java/tech/relaycorp/gateway/App.kt
gnarea: > Right now this is being called on every app start, and not only as needed. So you're going to get new keypairs every time. The methods should check themselves if they are needed, or we should check before calling them.

Good catch! We certainly don't want that.

> Also, they won't run in instrumented tests, which I'm not sure that's what we want.  We don't want to run the registration, because that requires other apps in the phone, but generating the keypair I think we want to.

Good point. Maybe the "gateway bootstrap" should be a suspend function like:

```
suspend fun bootstrapGateway() {
    localConfig.generateKeyPair()
    localConfig.generateCargoDeliveryAuth()
    if (mode != Mode.Test) {
        registerGateway.registerIfNeeded()
    }
}
```

And it could then be called from within a `ioScope.launch {}` block in `App.kt`, and also instrumentation tests. And this would be done in such a way that the keys are generated before attempting to register with the public gateway.

Thoughts?

sdsantos: The `suspend` makes sense. You might want to add the `@VisibleForTesting` annotation, because it should be a private method, but we want to use it for tests too.

But do note that, although we only want to generate a keypair once, we want to call registerIfNeeded at every app start.

app/src/androidTest/java/tech/relaycorp/gateway/background/endpoint/EndpointPreRegistrationServiceTest.kt
```diff
@@ -43,6 +43,7 @@ class EndpointPreRegistrationServiceTest {
     fun setUp() {
         AppTestProvider.component.inject(this)
         runBlocking {
+            localConfig.bootstrap()
```
gnarea, line 46: @sdsantos, one of the tests here (`requestPreRegistration`) is failing intermittently because the gateway key pair changes midway through the test because `App.bootstrapGateway()` gets called, so we end up calling `localConfig.bootstrap()` twice. I've just added some logs to verify this.

I'm only adding this call to `localConfig.bootstrap()` because, if I don't, I'd get an exception like this:

```
2021-01-20 17:07:01.992 16499-16535/tech.relaycorp.gateway E/AndroidRuntime: FATAL EXCEPTION: DefaultDispatcher-worker-2
    Process: tech.relaycorp.gateway, PID: 16499
    java.lang.RuntimeException: No key pair was found
        at tech.relaycorp.gateway.domain.LocalConfig.getKeyPair(LocalConfig.kt:31)
        at tech.relaycorp.gateway.domain.endpoint.EndpointRegistration.authorize(EndpointRegistration.kt:29)
        at tech.relaycorp.gateway.background.endpoint.EndpointPreRegistrationService.reply(EndpointPreRegistrationService.kt:73)
    ...
```

How can I make sure that `App.bootstrapGateway()` gets called before the tests run?

sdsantos: Usually a way for injecting the scope is ideal, it should already be possible in all classes using Dispatchers.IO, but I didn't design for it from the start.

If do for `waitForCheck` approach, I would say you can remove `localConfig.bootstrap()`, cause it will run anyway in the App class, you just need to wait for it before binding any service.

sdsantos (approved):

sdsantos:
Think I finally got it. Check it out and merge if all's ok @gnarea

gnarea:
Thanks! LGTM

PR #557, opened 2022-02-10
chore: Update Awala lib to use CertificateStore

sdsantos:


app/src/main/java/tech/relaycorp/gateway/domain/LocalConfig.kt
```diff
@@ -74,11 +93,10 @@ class LocalConfig
...
     }
 
-    private suspend fun generateCargoDeliveryAuth(idKeyPair: IdentityKeyPair) {
-        val cda = selfIssueCargoDeliveryAuth(
-            idKeyPair.privateKey,
-            idKeyPair.certificate.subjectPublicKey
-        )
+    private suspend fun generateCargoDeliveryAuth() {
+        val key = getIdentityKey()
+        val certificate = getIdentityCertificate() ?: return
```
gnarea, line 98: Is that `return` ever executed? I think `getIdentityCertificate()` always returns a value.

app/src/androidTest/java/tech/relaycorp/gateway/data/disk/AndroidPrivateKeyStoreTest.kt
```diff
@@ -8,37 +8,20 @@ import tech.relaycorp.awala.keystores.file.FileKeystoreRoot
...
         val store = AndroidPrivateKeyStore(root, androidContext)
 
-        store.saveIdentityKey(privateKey, certificate)
-        val retrievedKeypair = store.retrieveIdentityKey(certificate.subjectPrivateAddress)
-        assertEquals(privateKey, retrievedKeypair.privateKey)
-        assertEquals(certificate, retrievedKeypair.certificate)
-    }
-
-    @Test
-    fun overrideKey() = runBlockingTest {
```
gnarea, line 31: IIRC, I had to add this test after finding out the hard way that I had to do things differently in the unit under test. If we remove it, we risk making the same mistake if we were to refactor the unit under test later.

app/src/main/java/tech/relaycorp/gateway/common/Keys.kt
```diff
@@ -0,0 +1,8 @@
+package tech.relaycorp.gateway.common
+
+import tech.relaycorp.relaynet.wrappers.deserializeRSAKeyPair
+import java.security.PrivateKey
+import java.security.PublicKey
+
+fun PrivateKey.toPublicKey(): PublicKey =
+    encoded.deserializeRSAKeyPair().public
```
gnarea, line 8: Didn't we define a function somewhere to do this without serialising + deserialising? I'm pretty sure we we discussed it on Slack a few weeks ago. I can't find it though 🤔

gnarea (changes requested):
I think CI is failing because ktlint is complaining.

app/src/main/java/tech/relaycorp/gateway/common/Keys.kt
sdsantos: Found it, it was for the `privateAddress` method. But we'll need to import BounceCastle now, because we didn't expose that internal code.

PR #564, opened 2022-02-17
feat: Renew Identity certificates over the Internet

### LINKED ISSUE #507: Renew identity certificate over the Internet
When the Internet is available and the current identity certificate is due to expire in the next <= 90 days, we should re-register with the public gateway using the existing key pair.

See:

https://github.com/relaycorp/relaynet-gateway-android/blob/d43d4df564f6721daa4ada3668cd0a8f546251b9/app/src/main/java/tech/relaycorp/gateway/domain/publicsync/RegisterGateway.kt#L55-L56

# What to do with old certificates

Old certificates should continue to be used in all signature verification operations, for as long as the certificates are valid. Namely:

- All calls to `parcel.validate()` and `cargo.validate()`. The old (yet valid) certificates should be passed as `trustedCertificates`, along with the new certificate.
- The WebSocket-based, parcel collection endpoint in the PDC server: https://github.com/relaycorp/relaynet-gateway-android/blob/960f9467c812cbb70975e66e63ab980882554b00/app/src/main/java/tech/relaycorp/gateway/pdc/local/utils/ParcelCollectionHandshake.kt#L43 The old and new certificates should be included in the `trustedCertificates`.

Old certificates should only be deleted after they expire (e,g., once a day, when the app starts). However, expired certificates must not be included in the `trustedCertificates` above.

# What to do with the new certificate

The new certificate should be used in all signature verification operations (see above), as well as signature generation operations (i.e., `parcel.serialize()`, `cargo.serialize()`, `issueEndpointCertificate()`).

This shouldn't require code any further code changes, but I'm documenting this requirement to make sure we don't change the code inadvertently.

# Related issues

- #508 is the offline counterpart.
- #539

sdsantos:
Closes #507

### DIFF
--- a/app/src/main/java/tech/relaycorp/gateway/App.kt
+++ b/app/src/main/java/tech/relaycorp/gateway/App.kt
@@ -73,9 +73,13 @@ open class App : Application() {
         enqueuePublicSyncWorker()
 
         setupStrictMode()
-        bootstrapGateway()
-        startPublicSyncWhenPossible()
-        deleteExpiredCertificates()
+
+        backgroundScope.launch {
+            bootstrapGateway()
+            startPublicSyncWhenPossible()
+            deleteExpiredCertificates()
+        }
+
         registerActivityLifecycleCallbacks(foregroundAppMonitor)
     }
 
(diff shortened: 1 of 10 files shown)


app/src/main/java/tech/relaycorp/gateway/App.kt
```diff
@@ -138,6 +139,12 @@ open class App : Application() {
         }
     }
 
+    protected open fun deleteExpiredCertificates() {
+        backgroundScope.launch {
+            localConfig.bootstrap()
```
gnarea, line 144: Wrong method. I think it'd be good to add a test for this.

app/src/main/java/tech/relaycorp/gateway/domain/LocalConfig.kt
```diff
@@ -106,6 +106,10 @@ class LocalConfig
         fileStore.store(CDA_CERTIFICATE_FILE_NAME, cda.serialize())
     }
 
+    suspend fun deleteExpiredCertificate() {
```
gnarea, line 109: Nitpick: How about `deleteExpiredCertificates()`?

gnarea (changes requested):

app/src/main/java/tech/relaycorp/gateway/App.kt
```diff
@@ -75,6 +75,7 @@ open class App : Application() {
         setupStrictMode()
         bootstrapGateway()
         startPublicSyncWhenPossible()
+        deleteExpiredCertificates()
```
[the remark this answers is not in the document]
sdsantos: Yeah, should be sequentially 👍

app/src/main/java/tech/relaycorp/gateway/App.kt
sdsantos: 🤦

sdsantos: Unfortunately, this is not a very easy test to write. App instances are weird beasts. The easiest way would be to move all these initializations to another class that gets called by the App. But that probably is a better fit for another task.

gnarea (approved):
Thanks!

PR #569, opened 2022-02-23
feat: Rotate certificate from courier

### LINKED ISSUE #508: Save incoming id certificate renewals received via couriers
Only reason not to save it is if we already have a newer certificate.

Note that we should also replace our copy of the public gateway's id certificate as it may have changed too.

## Related issues

- This issue is blocked by https://github.com/relaycorp/relaynet-internet-gateway/issues/675
- #507 is the online counterpart to this issue.
- #539

sdsantos:
Closes #508

### DIFF
--- a/app/src/main/java/tech/relaycorp/gateway/domain/courier/RotateCertificate.kt
+++ b/app/src/main/java/tech/relaycorp/gateway/domain/courier/RotateCertificate.kt
@@ -1,0 +1,34 @@
+package tech.relaycorp.gateway.domain.courier
+
+import tech.relaycorp.gateway.common.Logging.logger
+import tech.relaycorp.gateway.data.preference.PublicGatewayPreferences
+import tech.relaycorp.gateway.domain.LocalConfig
+import tech.relaycorp.relaynet.messages.CertificateRotation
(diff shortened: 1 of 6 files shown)


app/src/main/java/tech/relaycorp/gateway/domain/courier/RotateCertificate.kt
```diff
@@ -0,0 +1,34 @@
...
+
+        val currentIdCert = localConfig.getIdentityCertificate()
+        val newIdCert = certRotation.subjectCertificate
+
+        if (currentIdCert.expiryDate >= newIdCert.expiryDate) return
+
+        localConfig.setIdentityCertificate(newIdCert)
+        certRotation.chain.first().let { publicGatewayCert ->
+            publicGatewayPreferences.setCertificate(publicGatewayCert)
+        }
```
gnarea, line 32: FYI, I think we should leave this as is in this PR, but I created two separate issues to simplify this as part of the maintenance:

- https://github.com/relaycorp/awala-jvm/issues/223
- https://github.com/relaycorp/relaynet-gateway-android/issues/570

gnarea (approved):
LGTM. Thanks!

Issue #571, opened 2022-02-25
Cargo Delivery Authorisation issuer isn't rotated


gnarea:
I forgot to call this out explicitly in the issues: We should rotate the self-issued certificate used in Cargo Delivery Authorisations.

Right now, we can only have exactly one as it's stored in a file that's generated the first time the app starts (and there's no rotation):
https://github.com/relaycorp/relaynet-gateway-android/blob/809ebbe4452375efb2514d146bb5b5eb86fecbfe/app/src/main/java/tech/relaycorp/gateway/domain/LocalConfig.kt#L116

Instead, we should do what we've been doing with the other certificates:

- Generate a new one when the latest one is about to expire.
- Use the latest certificate in all signature-producing operations. In this case, only here:
  https://github.com/relaycorp/relaynet-gateway-android/blob/8f01a2b0895c173a422b52da0d1e8bbd26c4e176/app/src/main/java/tech/relaycorp/gateway/domain/courier/GenerateCCA.kt#L27-L32
- Use all the valid certificates in all signature-verification operations. In this case, only here:
  https://github.com/relaycorp/relaynet-gateway-android/blob/c56cc7147c006c43e9d951e1d7814610ae8ba705/app/src/main/java/tech/relaycorp/gateway/data/disk/CargoStorage.kt#L30
- Delete all expired certificates.

We probably want to reuse our `FileCertificateStore` and introduce a second instance to manage CDA issuers only. I think that'd require changing the `FileCertificateStore` class to take a path prefix (under the existing root), so that we avoid overriding files.

sdsantos:
Tasks:

- [ ] Create second instance of `FileCertificateStore` for storing CDAs
- [ ] Switch from regular FileStore to FileCertificateStore for storing CDAs
- [ ] Generate new CDA when we don't have any, or the one we have stored is about to expire
- [ ] Method to return the latest valid CDA and use it in all signature-producing operations
- [ ] Method to return all valid CDAs and use it in all signature-verification operations
- [ ] Delete all expired certificates on app start

PR #588, opened 2022-03-28
fix: Update CertificateStore to use scope

### LINKED ISSUE #570: Avoid storing the public gateway's identity certificate
Private gateways don't need the identity certificate of their public peer, but we're still storing as a "leftover" from the days when we used to encrypt cargo with that certificate (before integrating the Channel Session Protocol).

The only thing we need from that certificate is its public key, so that we can:

- Issue Cargo Delivery Authorisations (CDAs)
- Compute the public gateway's private address

sdsantos:
Related with #571 
Closes #570

### DIFF
--- a/app/src/main/java/tech/relaycorp/gateway/data/preference/PublicGatewayPreferences.kt
+++ b/app/src/main/java/tech/relaycorp/gateway/data/preference/PublicGatewayPreferences.kt
@@ -14,7 +14,9 @@ import tech.relaycorp.gateway.data.model.RegistrationState
 import tech.relaycorp.relaynet.wrappers.x509.Certificate
 import javax.inject.Inject
 import javax.inject.Provider
+import javax.inject.Singleton
 
+@Singleton
 class PublicGatewayPreferences
 @Inject constructor(
     private val preferences: Provider<FlowSharedPreferences>,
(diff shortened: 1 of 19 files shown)


app/src/main/java/tech/relaycorp/gateway/data/preference/PublicGatewayPreferences.kt
```diff
@@ -56,6 +58,7 @@ class PublicGatewayPreferences
 
     suspend fun setCertificate(value: Certificate) {
         certificate.setAndCommit(Base64.encodeToString(value.serialize(), Base64.DEFAULT))
```
gnarea, line 60: I think it'd be best not to stop the certificate any more now that we're integrating this change. Instead of storing the certificate, I think we should only store its `subjectPrivateAddress`.

The original version of this app stored the public gateway's certificate because that was also used for encryption. However, now that the Channel Session Protocol is integrated, private gateways need not store their public gateways' certificates.

(The public gateway will always include its certificate when registering a private gateway, but that's mainly so that the private gateway can log or store specific elements of the certificate, and so that the private gateway can verify that its new certificate was issued by the public gateway)

gnarea (changes requested):

gnarea (dismissed):
Nice one. Thanks!

gnarea (dismissed):

gnarea (approved):

Issue #556, opened 2022-02-08
Improvement: alternative to Identity Certificate falling back to self-issued certificate


sdsantos:
# Describe the problem

Sometimes the Gateway app does not have an Identity Certificate available. For example, while the initial registration with the public gateway wasn't concluded, or a Certificate expires and we haven't been able to renew it. Right now we fall back to a self-signed certificate, but that only works to communicate with the Public Gateway, and not for the rest of the uses the Identity Certificate has.

# Describe the solution you'd like

Instead of just falling back to and storing a self-signed certificate, we should just create one for registration. And all other code that requires an Identity Certificate should be able to handle not having one (usually will mean not doing anything). Maybe we could even communicate to the user that the Gateway is not able to function fully until it registers again.

gnarea:
We should also get rid of our hardcoded copy of `frankfurt.relaycorp.cloud`'s certificate when we fix this:

https://github.com/relaycorp/relaynet-gateway-android/blob/2beac8a327e237683bff94371b870c7c93e14bff/app/src/main/java/tech/relaycorp/gateway/data/preference/PublicGatewayPreferences.kt#L50

PR #593, opened 2022-04-05
feat: Renew Cargo Delivery Authorisation

### LINKED ISSUE #571: Cargo Delivery Authorisation issuer isn't rotated

sdsantos:
Closes #571

### DIFF
--- a/app/src/main/java/tech/relaycorp/gateway/domain/LocalConfig.kt
+++ b/app/src/main/java/tech/relaycorp/gateway/domain/LocalConfig.kt
@@ -16,10 +16,9 @@ import tech.relaycorp.relaynet.wrappers.x509.Certificate
 import java.security.KeyPair
 import java.security.PrivateKey
 import java.security.PublicKey
-import java.time.ZonedDateTime
 import javax.inject.Inject
 import javax.inject.Provider
-import kotlin.time.days
+import kotlin.time.Duration.Companion.days
 import kotlin.time.toJavaDuration
 
 class LocalConfig
@@ -69,9 +68,7 @@ class LocalConfig
     }
 
     private suspend fun generateIdentityCertificate(privateKey: PrivateKey): Certificate {
-        val certificate = selfIssueCargoDeliveryAuth(
-            privateKey, privateKey.toPublicKey(), nowInUtc().plusYears(1)
-        )
+        val certificate = selfIssueCargoDeliveryAuth(privateKey, privateKey.toPublicKey())
         setIdentityCertificate(certificate)
         return certificate
     }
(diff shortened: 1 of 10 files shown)


app/src/main/java/tech/relaycorp/gateway/domain/LocalConfig.kt
```diff
@@ -65,50 +69,71 @@ class LocalConfig
...
+        publicKey: PublicKey,
+        expiryDate: ZonedDateTime
     ): Certificate {
         return issueGatewayCertificate(
             subjectPublicKey = publicKey,
             issuerPrivateKey = privateKey,
             validityStartDate = nowInUtc()
                 .minus(CalculateCRCMessageCreationDate.CLOCK_DRIFT_TOLERANCE.toJavaDuration()),
-            validityEndDate = nowInUtc().plusYears(1)
+            validityEndDate = expiryDate
```
gnarea, line 125: This doesn't need to match the `expiryDate` of the `getIdentityCertificate()` result. In fact, I think this should remain a constant value -- otherwise this would be one more place to change when we implement #556.

However, I think a year (as it was originally) is too long in hindsight. I think it should be 180 days for consistency with the desktop gateway -- although this, like all thresholds and limits in Awala, is mostly arbitrary until we have real-world usage. Thoughts?

app/src/main/java/tech/relaycorp/gateway/domain/LocalConfig.kt
```diff
@@ -118,9 +143,10 @@ class LocalConfig
     private suspend fun getPublicGatewayPrivateAddress() =
         publicGatewayPreferences.getPrivateAddress()
 
-    // Helpers
+    private fun Certificate.isExpiringSoon() =
+        expiryDate < (nowInUtc().plusNanos(CERTIFICATE_EXPIRING_THRESHOLD.inWholeNanoseconds))
 
     companion object {
-        internal const val CDA_CERTIFICATE_FILE_NAME = "cda_local_gateway.certificate"
+        private val CERTIFICATE_EXPIRING_THRESHOLD = 1.days
```
gnarea, line 150: This should be much longer, since this is basically the length of time we're willing to wait for incoming cargo: If `isExpiringSoon()` is called when the certificate has 1 day 10 minutes left, that'd mean the incoming cargo must've been received in less than 1 day 10 minutes.

I think 90 days (half of the initial TTL of 180 days) would be good for consistency per the above.

gnarea (changes requested):
Obrigado! 👍🏾 Left a couple of comments

gnarea (approved):
🙇🏾‍♂️ 🚀

PR #713, opened 2023-10-11
Update awala dependencies and kotlin to 1.9

### LINKED ISSUE #711: App frequently disconnects from the Internet gateway
We experience this too often whilst testing Letro: Outgoing messages are always sent, but incoming messages are not received. 

When this happens and we check the Awala app, it normally says that we're disconnected from the Internet, but it's not uncommon to see that we're connected to Awala via the Internet. Either way, the logs don't show anything relevant.

I suspect the PoWeb library may be the culprit, but we're not seeing its logs due to #709.

sdsantos:
Closes #711


app/jacoco.gradle
```diff
@@ -44,13 +44,13 @@ task jacocoAndroidTestReport(type: JacocoReport) {
     if (project.hasProperty('codeCoverageDataLocation')) {
         executionData.from += fileTree(dir: codeCoverageDataLocation, includes: ['**/*.ec'])
     }
-
+\
```
gnarea, line 47: Typo?

gnarea (changes requested):
LGTM, but I think I spotted a typo / syntax error.

gnarea (approved):
Thanks!

gnarea (approved):

PR #718, opened 2023-10-30
fix: Register gateway on sync

### LINKED ISSUE #714: Connecting to the Internet after opening the app for the first time doesn't trigger registration with Internet gateway
# Describe the bug

If you're disconnected from the Internet when you open the gateway app for the first time, you'll get the screen saying that you're disconnected from the Internet (as expected). However, if you then connect to the Internet, the screen is updated to say that you're connected, but there's no attempt to actually register with the Internet gateway, so no Awala-compatible app can use the local gateway app.

# How to reproduce

1. Clear the Awala app data, or reinstall the app.
2. Disconnect from the Internet.
3. Open the Awala app and go through the onboarding screens.
4. Observe the screen that says that you're disconnect from Awala.
5. Connect to the Internet.
6. Observe the screen that says that you're connected to Awala.
7. Use Letro or Awala Ping. The Awala SDK will throw an exception because the Awala app isn't yet registered with its Internet gateway.

# Expected behavior

The Awala app tries to register with the Internet gateway as soon as the connection with the Internet is established.

# Additional context

Killing the Awala app after connecting to the Internet is a workaround.

sdsantos:
Closes #714

Also fixes 2 issues I found:
- Certificates weren't deleted if expired, because that method was called after the never-ending public sync method
- The gateway certificate we received expires after 30 days, and we were always renewing them if the expiration date was less than 90 days (changed to 25 days)

### DIFF
--- a/app/src/main/java/tech/relaycorp/gateway/App.kt
+++ b/app/src/main/java/tech/relaycorp/gateway/App.kt
@@ -23,7 +23,6 @@ import tech.relaycorp.gateway.common.di.AppComponent
 import tech.relaycorp.gateway.common.di.DaggerAppComponent
 import tech.relaycorp.gateway.domain.LocalConfig
 import tech.relaycorp.gateway.domain.publicsync.PublicSync
-import tech.relaycorp.gateway.domain.publicsync.RegisterGateway
 import java.security.Security
 import java.time.Duration
 import java.util.logging.Level
@@ -57,9 +56,6 @@ open class App : Application() {
     lateinit var localConfig: LocalConfig
 
     @Inject
-    lateinit var registerGateway: RegisterGateway
-
-    @Inject
     lateinit var publicSync: PublicSync
 
     @Inject
@@ -77,8 +73,8 @@ open class App : Application() {
 
         backgroundScope.launch {
             bootstrapGateway()
-            startPublicSyncWhenPossible()
-            deleteExpiredCertificates()
+            launch { startPublicSyncWhenPossible() }
+            launch { deleteExpiredCertificates() }
         }
 
         registerActivityLifecycleCallbacks(foregroundAppMonitor)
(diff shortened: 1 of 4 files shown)


gnarea:
Thanks! 

The PR LGTM but I won't be able to test it today and I want to check a few different scenarios, so I'll approve/merge it tomorrow.

> The gateway certificate we received expires after 30 days

Are you sure? I can investigate this tomorrow, but I had a quick check and [we're setting it](https://github.com/relaycorp/relaynet-internet-gateway/blob/06927bbe9d94d177cb32dc4a2716bb0d5b4fbded/src/pki.ts#L74) to [6 months](https://github.com/relaycorp/relaynet-internet-gateway/blob/06927bbe9d94d177cb32dc4a2716bb0d5b4fbded/src/pki.ts#L21), so if the certificate you get expires after 30 days, then that's a bug in the Internet Gateway or the Awala JVM lib.

sdsantos:
> > The gateway certificate we received expires after 30 days
> 
> Are you sure? I can investigate this tomorrow, but I had a quick check and [we're setting it](https://github.com/relaycorp/relaynet-internet-gateway/blob/06927bbe9d94d177cb32dc4a2716bb0d5b4fbded/src/pki.ts#L74) to [6 months](https://github.com/relaycorp/relaynet-internet-gateway/blob/06927bbe9d94d177cb32dc4a2716bb0d5b4fbded/src/pki.ts#L21), so if the certificate you get expires after 30 days, then that's a bug in the Internet Gateway or the Awala JVM lib... And we shouldn't change that here because it really has to be valid for a few months.

Just tested again on a fresh install and the `privateNodeCertificate` we get back inside the `PrivateNodeRegistration` is expiring at `2024-01-26T20:56:37Z[Europe/Lisbon]`.

gnarea:
It's a bug in the Internet Gateway: I wasn't being alerted to the fact that [the daily certificate rotation job is failing](https://github.com/relaycorp/cloud-gateway/issues/93), so the last certificate that was successfully generated expires on 26th Jan 2024... Which means that all the certificates it issues will have that expiry date too.

So, can you please undo the change you made to renew the certificate every 25 days?

sdsantos:
@gnarea reverted.

gnarea (approved):
Thanks!

PR #720, opened 2023-11-06
fix: Improve connection state checks speed

### LINKED ISSUE #717: App doesn't try to reconnect to its Internet gateway
(Sorry about the vague title, but I don't really know how to summarise it, or even if it's one or more issues.)

I was using Letro this afternoon until 18:00 via WiFi, and everything worked fine. Then I left that place and switched to 5G.

At 20:56, connected to another WiFi network, I open Letro and send a message, but nothing happens. So I open Awala and see the usual "You're disconnected from Awala" screen, so I checked logcat and here's what I found:

- Letro: The message was indeed sent to Awala. Interestingly, there are no [logs saying that we're setting up the Awala SDK](https://github.com/relaycorp/letro-android/blob/3378ae6ad2fbad2ddc8f32fe86b59302c05df7b4/app/src/main/java/tech/relaycorp/letro/awala/AwalaManager.kt#L121), so I'm guessing `Awala.setup()` wasn't called again at 20:56, but I can see the log entry from a few hours ago. Unfortunately, as I was writing this logcat lost the logs prior to 20:56, but here's [letro-logs.txt](https://github.com/relaycorp/relaynet-gateway-android/files/13182537/letro-logs.txt).
- Awala.  There was no [log saying that it was delivering the parcel to the Internet gateway](https://github.com/relaycorp/relaynet-gateway-android/blob/bf0b6bd2dd8e4cd6a0daea542db206d5d70cfc48/app/src/main/java/tech/relaycorp/gateway/domain/publicsync/DeliverParcelsToGateway.kt#L92), and I also noticed that the last public sync (which was old) was done with `keepAlive=false` -- unfortunately, logcat lost these logs as I was writing this.

I triggered a network change event by switching from WiFi to 5G, and then the app reported that it was connected to the Internet. When I checked the [awala-logs.txt](https://github.com/relaycorp/relaynet-gateway-android/files/13182627/awala-logs.txt), I found that it'd just delivered the parcel from Letro.

So, two questions:

- Why didn't that `OutgoingMessage.send()` call from the SDK didn't cause the app to send the parcel to the Internet gateway?
- Why wasn't this app running the public sync with `keepAlive=true` whilst I was using Letro?

I wonder if the problem is that the last time we called `GatewayClient.bind()` was a few hours ago. If so, is there anything we can do in the SDK automatically?

sdsantos:
Fixes #717

We're repeating the connection state check every 5 seconds. But the 3 checks themselves did not have timeouts, so that meant it could take several minutes just to complete one check. I've now capped each check to 5 seconds as well, so worse case scenario it could take 20 seconds between connection state checks.

### DIFF
--- a/app/src/main/java/tech/relaycorp/gateway/background/PingRemoteServer.kt
+++ b/app/src/main/java/tech/relaycorp/gateway/background/PingRemoteServer.kt
@@ -8,11 +8,14 @@ import io.ktor.client.request.head
 import io.ktor.network.selector.ActorSelectorManager
 import io.ktor.network.sockets.aSocket
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.TimeoutCancellationException
+import kotlinx.coroutines.withTimeout
 import tech.relaycorp.gateway.BuildConfig
 import tech.relaycorp.gateway.common.Logging.logger
 import java.io.IOException
 import java.util.logging.Level
 import javax.inject.Inject
+import kotlin.time.Duration.Companion.seconds
 
 class PingRemoteServer
 @Inject constructor() {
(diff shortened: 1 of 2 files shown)


gnarea (approved):
Thanks! LGTM. Will keep an eye on this... hopefully this will be the last fix to this connectivity issues. 🤞🏾

PR #726, opened 2024-01-04
chore: Avoid hardcoded internet gateway certificate

### LINKED ISSUE #556: Improvement: alternative to Identity Certificate falling back to self-issued certificate

sdsantos:
Closes #556

Unless I'm mistaken, we don't need to even a self-issued Internet Gateway certificate, we only need our local self-issued certificate.

I tried a simple approach:
- returning `null`s while we don't have the Internet Gateway certificate and ID
- Skip operations that need the Internet Gateway certificate and ID with relevant logging

I haven't changed the UI/UX to show something while the Gateway is not registered. That would need maybe a new illustration? Or at least new copy for the home screen. Should that be a another task?


gnarea (changes requested):
The bulk of the change appears to be handling the case where the id or public key of the Internet gateway is missing. However, unless I'm missing something, that's different from #556 (i.e., the private gateway's certificate is missing).

Having said this, I forgot to update #556 with an important detail: whist we should indeed remove `public_gateway_cert.der`, we should replace it with the [_connection parameters_](https://docs.relaycorp.tech/awala-jvm/awala/tech.relaycorp.relaynet/-node-connection-params/index.html) of `frankfurt.relaycorp.tech`* just [like we do in Awala Ping](https://github.com/relaycorp/awala-ping-android/blob/5c665d0d56db7aa024fbb0dda93e17ffb17f93ad/app/src/main/res/raw/default_public_peer_connection_params.der). This `NodeConnectionParams` instance will contain the `internetAddress` (`frankfurt.relaycorp.tech`) and `publicKey` of the Internet gateway -- meaning that we can populate the `InternetGatewayPreferences` with those details as soon as the app starts for the first time.

Unfortunately, the current version of the Awala Internet Gateway doesn't yet expose this file like the Awala Pong server does, but I'll implement it today or tomorrow. I'll ping you once I've done that.

As for #556, the goal is to properly handle the scenario where we haven't yet got an identity certificate issued by our Internet gateway (the one we get after [successfully registering with the Internet gateway](https://github.com/relaycorp/relaynet-gateway-android/blob/82afd23665975be10c9305d218098e86948175f6/app/src/main/java/tech/relaycorp/gateway/domain/publicsync/RegisterGateway.kt#L88)).

The problem we have today is that we're faking it with a self-issued certificate that has no validity on the network. For example, when an app registers an endpoint before we have a valid identity certificate, we generate a self-issued certificate so that we can subsequently issue a certificate for the endpoint -- instead, we should block the registration of the endpoint (with the "gateway not ready" approach we have today for other things, but I couldn't find the respective code to link to it).

sdsantos:
@gnarea But we're already handling endpoint registration when the gateway is not registered yet: https://github.com/relaycorp/relaynet-gateway-android/blob/82afd23665975be10c9305d218098e86948175f6/app/src/main/java/tech/relaycorp/gateway/background/endpoint/EndpointPreRegistrationService.kt#L67-L70

Maybe that's what threw me off track in the task?

gnarea:
Hi @sdsantos 👋🏾 

Good catch. I think the issue is partly out of date then: we're no longer using any self-issued certificate that should be removed. Whilst we're indeed handling the case the scenario where the private gateway has never registered, we're not handling the case where its certificate expired (`getRegistrationState()` only seems to consider whether we ever registered). Obviously, the latter isn't actually a self-issued certificate, it's just expired, but it's invalid too.

On the other hand, implementing the connection params took a lot longer than I anticipate but it's now available in the new environment that replaces Frankfurt: `belgium.relaycorp.services` (#728). Its connection parameters file can be downloaded from:

```
https://belgium-poweb.relaycorp.services/connection-params.der
```

So we basically have to:

- Refuse endpoint pre-registrations with a `GATEWAY_NOT_REGISTERED` message when the private gateway's certificate has expired.
- Replace [the hard-coded Internet gateway certificate](https://github.com/relaycorp/relaynet-gateway-android/blob/753b5376eec955d76d987980ef372f35697feb78/app/src/main/res/raw/public_gateway_cert.der) with the connection params file above.

Please let me know if you have any questions!

Issue #724, opened 2023-12-27
Incoming messages are not received until the app is killed


gnarea:
# Describe the bug

This app is constantly failing to collect parcels from its Internet gateway, until we kill it or switch networks. There are no issues on the server as far as I can see.

We can reliably reproduce this within a few minutes of using Letro. It's basically #711 and #717 with a twist: Most of the time, the Awala app reports it **is** connected to the Internet (this is new since those two issues).

# How to reproduce

1. Install the latest version of [Letro from source](https://github.com/relaycorp/letro-android) on two phones.
2. Create an account on each phone.
3. Pair the two users.
4. Send messages to each other.

At some point between steps 2 and 4, things will get stuck due to this issue.

# Additional context

I typically test with a Pixel 2 (Android 11) and a Pixel 5 (Android 14), and this happens far more often on the Pixel 2.

sdsantos:
I am able to reproduce the issue quickly. What I've seen so far is:
- You need to kill the Gateway app specifically to get the messages received;
- We are still re-registering every time the public sync starts because we get a certificate that's only 30 days valid;
- Every 30 seconds we get a `WebSocket connection ended abruptly (null). Will retry.` log but even before getting the first one we don't receive parcels.
- I changed the code to trigger collect parcels with keepAlive=false (`PublicSync.syncOneOff`) and we get parcels fine.

Possible issues:
a) The InternetGateway is not honouring the keepAlive=true and only delivers the parcels it has once it starts
b) The PoWebClient or the CollectParcels code for receiving parcels has some bug that prevents from getting more than 1 parcel
c) We have a web socket connection issue that makes it fail very quickly and never recover, even with our retries

gnarea:
Thanks for looking into this Sergio!

> I changed the code to trigger collect parcels with keepAlive=false (PublicSync.syncOneOff) and we get parcels fine.

Just to clarify, do you mean that you changed that today in a local branch (not the remote `master`)? If so, did that appear to solve the problem consistently?

> The InternetGateway is not honouring the keepAlive=true and only delivers the parcels it has once it starts

If forcing `keepAlive=false` causes parcels to be retrieved reliably, then it does indeed sound like a bug on the Internet gateway. The weird thing then is that it does appear to happen a lot more often with the older phones (Nexus 5X/Android 8 and Pixel 2/Android 11), than the newer ones (Pixel 5 and 8 with Android 14).

sdsantos:
[...]

PR #730, opened 2024-02-20
fix: Don't do network calls when apps are in background

sdsantos:
Addresses https://github.com/relaycorp/relaynet-gateway-android/issues/724

The other task needed to close this ticket is to have endpoint apps bind and unbind from the Gateway when they are in the foreground/background.


gnarea (approved):
Thanks! LGTM, but just in case let's wait for the endpoint/app changes before merging this.

PR #732, opened 2024-02-29
fix: Update Internet GW default params and fix wrong identity certificate usage

sdsantos:
Related with https://github.com/relaycorp/relaynet-gateway-android/pull/726/files


app/src/main/res/raw/public_gateway_cert.der
gnarea, line 1: I'd suggest to rename this to something like `default_internet_gateway_params.der`, as this is no longer a certificate.

app/src/main/java/tech/relaycorp/gateway/domain/LocalConfig.kt
```diff
@@ -62,6 +62,14 @@ class LocalConfig
     suspend fun setIdentityCertificate(
         leafCertificate: Certificate,
         certificateChain: List<Certificate> = emptyList(),
+    ) {
+        certificateStore.get().delete(leafCertificate.subjectId, getInternetGatewayId())
+        addIdentityCertificate(leafCertificate, certificateChain)
+    }
+
+    suspend fun addIdentityCertificate(
```
gnarea, line 70: I think the combination of  `setIdentityCertificate()` + `addIdentityCertificate()` is going to make it more confusing which cert we're talking about.

Also, I just realised that the whole logic in the original `setIdentityCertificate()` + `generateIdentityCertificate()` is wrong: https://github.com/relaycorp/relaynet-gateway-android/blob/e02d0ba460d6af9f9d15b48d4a60ff65bbcab2ac/app/src/main/java/tech/relaycorp/gateway/domain/LocalConfig.kt#L62-L77. We're basically self-issuing a certificate for the courier communication, but when we save it we store it as if it'd been issued by the Internet gateway.

With the above in mind, I think we should reflect it in the code that the private gateway should have exactly two certificates. For example, we could replace `setIdentityCertificate()` + `addIdentityCertificate()` with the following functions, and we wouldn't have to delete any certificates:

```kotlin
    // This is the one we should call when we register with an Internet gateway
    suspend fun setParcelDeliveryCertificate(
        privateGatewayCertificate: Certificate,
        certificateChain: List<Certificate> = emptyList(),
    ) {
        certificateStore.get()
            .save(
                CertificationPath(privateGatewayCertificate, certificateChain),
                getInternetGatewayId(), // or: privateGatewayCertificate.issuerId
            )
    }


    // This one replaces the original setIdentityCertificate()
    suspend fun setCargoDeliveryCertificate(
        privateGatewayCertificate: Certificate,
    ) {
        certificateStore.get()
            .save(
                CertificationPath(privateGatewayCertificate, emptyList()),
                privateGatewayCertificate.subjectId
            )
    }
```

We should probably rename `generateIdentityCertificate()` to something like `generateCargoDeliveryCertificate() for consistency.

gnarea (changes requested):

gnarea (approved):
Thanks! 🙇🏾

